PR #1, opened 2025-10-30
Add member mailer

giverm:
## Ticket

Resolves TSS-311

## Changes

Adds MemberMailer with `exempt_email` and `action_required_email`. `CertificationCase#determind_ex_parte_exepmtion` to send an exempt email if the automated exemption check finds that the member is exempt, and send the action required email otherwise. The action required email was added to the exemption check, as this is currently the last step in the automated process. Once the "ex_parte_community_engagement_check" step has been implemented, the action required email should be moved.

## Context for reviewers

The emails can be previewed in the local dev environment at localhost:3000/rails/mailers

### Exempt Email
**HTML**
Screenshot 2025-11-04 at 1 21 44 PM
**Plain Text**
Screenshot 2025-11-04 at 1 22 01 PM

### Action Required Email
**HTML**
Screenshot 2025-11-04 at 2 04 17 PM

**Plain Text**
Screenshot 2025-11-04 at 2 04 27 PM

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/views/member_mailer/action_required_email.html.erb
+++ b/reporting-app/app/views/member_mailer/action_required_email.html.erb
@@ -2,11 +2,11 @@
 
 <p><%= t('.need_information') %></p>
 
-<p><%= t('.login_instructions', login_link: link_to(@login_url, @login_url), due_date: content_tag(:strong, @due_date)).html_safe %></p>
+<p><%= t('.login_instructions_html', login_link: link_to(@login_url, @login_url), due_date: content_tag(:strong, @due_date)).html_safe %></p>
 <ol>
   <li>
     <strong style="display: block; margin-bottom: 0.5em;"><%= t('.report_activities.title') %></strong>
-    <%= t('.report_activities.description').html_safe %>
+    <%= t('.report_activities.description_html').html_safe %>
     <%= t('.report_activities.helpful_examples') %>
     <ul>
       <li><%= t('.report_activities.work_income').html_safe %></li>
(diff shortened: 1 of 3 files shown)


reporting-app/spec/mailers/previews/member_mailer_preview.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class MemberMailerPreview < ActionMailer::Preview
+  def action_required_email
+    certification = mock_certification
+    MemberMailer.with(certification: certification).action_required_email
+  end
+
+  def exempt_email
+    certification = mock_certification
+    reasons = [ :age_exemption, :full_time_student ]
```
baonguyenNava: Here, we should change this to an array of strings as well.

baonguyenNava (changes requested):
There is a critical change necessary with how we pass `reasons`. Other than that, looks good.

sumiat:
Could we update the exemption email content? I think it could sound better to take more of the burden away from the user. Updated in the content in the doc: https://docs.google.com/document/d/1a2rcsUb9Dwj0e_G5MLXgRUPW1er2h6pavT9wgQ9sZZE/edit?tab=t.0#heading=h.d572cvvzzc42

reporting-app/spec/mailers/previews/member_mailer_preview.rb
giverm: removed

baonguyenNava (approved):

PR #5, opened 2025-11-03
Add pregnancy_status to Certification create and allow for automated exemption

baonguyenNava:
## Ticket

Resolves [TSS-421](https://linear.app/nava-platform/issue/TSS-421/add-ex-parte-scenario-data-pregnancy)

## Changes

Add prenancy_status to demo certifications create form
Add pregnancy status to automated ex_parte exemption process.

## Testing

Here are examples of how we are storying multiple reasons at the database level.
Screenshot 2025-11-03 at 1 34 34 PM
Screenshot 2025-11-03 at 1 34 42 PM

> Provide evidence that the code works as expected. Explain what was done for testing and the results of the test plan. Include screenshots, [GIF demos](https://www.cockos.com/licecap/), shell commands or output to help show the changes working as expected. ProTip: you can drag and drop or paste images into this textbox.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/certification_case.rb
+++ b/reporting-app/app/models/certification_case.rb
@@ -75,7 +75,8 @@ class CertificationCase < Strata::Case
 
         reasons = {
           age_under_19: :age_under_19_exempt,
-          age_over_65: :age_over_65_exempt
+          age_over_65: :age_over_65_exempt,
+          pregnant: :pregnancy_exempt
         }
 
         # TODO: Extract to Fact class that inherits from Strata::RulesEngine::Fact
--- a/reporting-app/app/models/rules/exemption_ruleset.rb
+++ b/reporting-app/app/models/rules/exemption_ruleset.rb
@@ -10,16 +10,16 @@ module Rules
       age < 19
     end
 
-    def pregnant(pregnancy_status)
+    def is_pregnant(pregnancy_status)
       return if pregnancy_status.nil?
 
       pregnancy_status
     end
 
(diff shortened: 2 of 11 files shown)


reporting-app/app/models/rules/exemption_ruleset.rb
```diff
@@ -10,10 +10,16 @@ def age_under_19(age)
       age < 19
     end
 
-    def eligible_for_age_exemption(age_under_19, age_over_65)
-      return if age_under_19.nil? && age_over_65.nil?
+    def pregnant(pregnancy_status)
```
MichaelCrawfordNava: Seems like this should be `pregnant?` with a question mark

reporting-app/app/models/certification_case.rb
```diff
@@ -75,7 +75,8 @@ def determine_ex_parte_exemption(eligibility_fact)
 
         reasons = {
```
baonguyenNava: I will refactor this whole block after I implement creating Determinations for accepting an exemption app form and an activity report app form.

MichaelCrawfordNava (approved):

reporting-app/app/models/rules/exemption_ruleset.rb
baonguyenNava: I renamed it to `is_pregnant`. We can't use `?` on the method names because of how the params are passed around. Also, the `determination_data` will store the method name `is_pregnant: true`, so using a ? would make it odd.

PR #17, opened 2025-11-07
Add optimized member status for batch lookups

baonguyenNava:
## Changes

Add `determine_many` to MemberStatusService. This will call bulk queries so we load all Certification, CertificationCase, and Determination objects at once in memory. This allows us to create a member status object for many Certifications with only three queries at most.

Add for_certifications scope

Add latest_per_subject scope - Grabs the latest determination per subject_id, selected on the most recent determination by created_at.

## Context for reviewers

Member Status was recently introduced. It performs two queries, one to retrieve either Certification or Certification Case, and another to retrieve the latest determination for that query. This is inefficient if we call MemberStatusService.determine within a mapping function as it results in an N+1 query.

## Testing

Would need to verify with testing with batch upload.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/determination.rb
+++ b/reporting-app/app/models/determination.rb
@@ -50,6 +50,18 @@ class Determination < Strata::Determination
 
   default_scope { order(created_at: :desc) }
 
+  # Batch query scopes
+  scope :for_certifications, ->(certification_ids) {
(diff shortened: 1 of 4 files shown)


giverm (approved):
Looks good overall!

PR #20, opened 2025-11-07
Update activities form month select

giverm:
## Ticket
Resolves TSS-417
Resolves TSS-354

## Changes
This commit updates the activities form to populate the months select with the reporting periods selected in a previous application step.

## Testing

Kapture 2025-11-07 at 15 43 00

♻️ Environment destroyed ♻️


reporting-app/app/controllers/activities_controller.rb
```diff
@@ -113,6 +117,15 @@ def set_activity_report_application_form
       @activity_report_application_form = ActivityReportApplicationForm.find(params[:activity_report_application_form_id])
     end
 
+    def selectable_months
```
baonguyenNava: I think this should be moved to the `ActivityReportApplicationForm` model. Then it would make create model specs for this method a lot easier.

PR #23, opened 2025-11-10
Show exemption approved when member status is exempt

baonguyenNava:
## Ticket

Resolves #[TSS-498](https://linear.app/nava-platform/issue/TSS-498/demo-fix-must-have-batch-and-api-exempt-users-see-exemption-screen)

## Changes

- added before action on controller to set member status based on certification '
- added logic to render exemption accepted page if status is Exempt

## Context for reviewers

If a member triggers an Exemption based on Ex Parte data, then the member should see exemption accepted page. Examples would be age under 19 or pregnancy. This evaluation is triggered as the first step in the business process on a certification case created. 

## Testing

image

Screenshot 2025-11-10 at 5 54 07 PM
Screenshot 2025-11-10 at 5 54 45 PM
2025-11-10 at 17 51 08 - Determination

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/helpers/dashboard_helper.rb
+++ b/reporting-app/app/helpers/dashboard_helper.rb
@@ -2,9 +2,9 @@
 
 module DashboardHelper
   def determine_dashboard_view
-    return "exemption_approved" if @member_status&.status == MemberStatus::EXEMPT
-
-    if @certification.nil?
+    if member_status_exempt?
+      "exemption_approved"
+    elsif @certification.nil?
       "no_certification"
     elsif are_activity_report_or_exemption_incomplete?
       "new_certification"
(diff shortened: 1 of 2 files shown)


reporting-app/app/helpers/dashboard_helper.rb
```diff
@@ -2,6 +2,8 @@
 
 module DashboardHelper
   def determine_dashboard_view
+    return "exemption_approved" if @member_status&.status == MemberStatus::EXEMPT
```
nkutub: I don't think this is working I dont see where MemberStatus::EXEMPT gets set in the code anywhere, I also do see a ExemptionDeterminationService that is called but I don't think its working , I set my age as low and it does not seem to be working

giverm (approved):
Looks good! I think it could also benefit from a request spec to illustrate the change.

PR #24, opened 2025-11-10
fix item icons on certification screener page

nkutub:
## Ticket

Resolves #{[TSS-494](TSS-494/demo-fix-nice-to-have-add-item-icons-on-the-tell-us-about-your)}

## Changes

> use path helper for icons on screener page.

## Context for reviewers

This was a simple fix to user the path helper in rails vs hardcoded paths. The icons will show locally on dev but when deployed, rails will fingerprint all images to bust cache. To fix we use the rails path helper that adds the fingerprinted path. 

## Testing
cant really test this locally, but local box still works 

image

♻️ Environment destroyed ♻️


giverm (approved):
looks good!

PR #28, opened 2025-11-12
Resolve form_with deprecation warning

giverm:
Passing `nil` to the `:model` argument in `form_with` has been deprecated and will raise an exception in later versions of Rails. https://github.com/rails/rails/pull/50931

This work resolves the deprecation warning when a user attempts to update their email address with one that is invalid. Previously, this would re-render the edit view without setting @password_form causing the warning to be displayed. Now, the Users::Accounts controller has a `set_password_form` callback to ensure the password_form is set and the edit view renders correctly.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/controllers/users/accounts_controller.rb
+++ b/reporting-app/app/controllers/users/accounts_controller.rb
@@ -2,11 +2,11 @@
 
 class Users::AccountsController < ApplicationController
   before_action :authenticate_user!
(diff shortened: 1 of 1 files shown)


giverm:
Opened a PR for this on navapbc/template-application-rails https://github.com/navapbc/template-application-rails/pull/136

giverm:
Closed in favor of https://github.com/navapbc/template-application-rails/pull/136

PR #39, opened 2025-11-19
fix update on certification request page

nkutub:
## Ticket

Resolves # [TSS-442/](https://linear.app/nava-platform/issue/TSS-442/bug-activity-report-flow-breaks-if-update-is-clicked-on-certification)
## Changes

- update the certification controller PATCH method to pull the nested params for certification requirements on updates. 

## Context for reviewers

When creating a new certification under demo, if a user updates the certification, the certification requirments get lost. this ultimately breaks the certification and will not work on other functions in the app. The change can been seen after update on the form, the requirements lose some values . 

## Testing
When clicking on update the values in the form fields all stay the same, nothing is lost.

image

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/controllers/certifications_controller.rb
+++ b/reporting-app/app/controllers/certifications_controller.rb
@@ -80,9 +80,6 @@ class CertificationsController < StaffController
           if cert_params[:certification_requirements].present? && cert_params[:certification_requirements].is_a?(String)
             parsed_requirements = JSON.parse(cert_params[:certification_requirements])
 
-            # Permit all Requirements and RequirementParams attributes, plus nested params hash
-            permitted_keys = (Certifications::Requirements.attribute_names | Certifications::RequirementParams.attribute_names).map(&:to_sym)
-
             cert_params[:certification_requirements] =
               ActionController::Parameters.new(parsed_requirements).permit(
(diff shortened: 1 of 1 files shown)


reporting-app/app/controllers/certifications_controller.rb
```diff
@@ -78,10 +78,17 @@ def certification_params
...
+            permitted_keys = (Certifications::Requirements.attribute_names | Certifications::RequirementParams.attribute_names).map(&:to_sym)
+
             cert_params[:certification_requirements] =
-              ActionController::Parameters.new(
-                JSON.parse(cert_params[:certification_requirements])
-              ).permit((Certifications::Requirements.attribute_names | Certifications::RequirementParams.attribute_names).map(&:to_sym))
+              ActionController::Parameters.new(parsed_requirements).permit(
+                *permitted_keys,
+                months_that_can_be_certified: [],
+                params: permitted_keys
```
giverm: After looking at the params more closely I think we could do something like:

```suggestion
                *Certifications::Requirements.attribute_names.map(&:to_sym).excluding(:months_that_can_be_certified, :params),
                months_that_can_be_certified: [],
                params: Certifications::RequirementParams.attribute_names.map(&:to_sym)
```

Then we can remove setting `permitted_keys` above

giverm (approved):

PR #79, opened 2025-12-04
Add user attributes for access control demo

### LINKED ISSUE #54: Authz: Update User model with access control attributes
## User Story
As a user
I want to access to only the things that I need access to
So that I can have a great user experience

## Acceptance Criteria
- A role, program, and region can be manually (via the DB) assigned to a user

## Technical Notes
- Tables to update:
  -  users:
      - role
      -  region
      - full_name

baonguyenNava:
## Ticket

Resolves #54 

## Changes

Add the following user attributes:
role - singular but we can revise to take multiple roles in the future
region

Validations are not included, to minimize impact of demo specific requirements on the rest of OSCER.

## Context for reviewers

Adding user attributes for use with demoing ABAC.

## Testing

I did a brief smoke test to ensure app still works on signing with a new user.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/user.rb
+++ b/reporting-app/app/models/user.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
-  attribute :full_name
-  attribute :region
-  attribute :role
+  attribute :full_name, :string
+  attribute :region, :string
+  attribute :role, :string
 
   devise :auth_service_authenticatable, :timeoutable
   attr_accessor :access_token
--- a/reporting-app/app/models/user.rb
+++ b/reporting-app/app/models/user.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
-  attribute :program, default: "Medicaid"
+  attribute :full_name
   attribute :region
   attribute :role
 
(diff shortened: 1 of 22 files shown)


reporting-app/app/models/user.rb
lorenyu: While it's not required in Rails I think we should have a practice of always defining `attribute` (or `strata_attribute` for all attributes, and not rely on the developer to need to look into the DB schema to know about attributes that are inherited from the table columns.

reporting-app/app/services/auth_service.rb
```diff
@@ -92,6 +92,11 @@ def create_db_user(uid, email, provider)
         uid: uid,
         email: email,
         provider: provider,
+        first_name: Faker::Name.first_name,
+        last_name: Faker::Name.last_name,
+        role: "caseworker",
+        program: "Medicaid",
+        region: "Northwest"
```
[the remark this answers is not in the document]
baonguyenNava: I can fix the region part. The fake names are to show in the UI that will be demo'ed. We won't be able to edit a user. Should I remove the names and not have names from this design? https://www.figma.com/design/hyfC70INCqz2VABg1TTK57/-TSS-512--Design--Supervisor-UI-for-role-assignment?node-id=1-4&t=9j1fIaSlHVzJ4qiG-1

reporting-app/app/models/user.rb
```diff
@@ -1,23 +1,19 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
+  attribute :full_name
+  attribute :region
+  attribute :role
```
lorenyu: worth defining the types? like `:string`

lorenyu (approved):

PR #81, opened 2025-12-04
Add rails runner instructions for updating user attributes

### LINKED ISSUE #56: Authz: Add/Remove role, program, region to User via Rake task
## User Story

As an app administrator

I want to set the follow user attributes for access control testing.
- role
- region
- full_name

## Acceptance Criteria
- A role can be added to a user via a Rake task
- A role can be removed from a user via a Rake task
- A region can be added for a user via a Rake task
- A region can be removed for a user via a Rake task
- Full name can be set for a user via a Rake task

## Technical Notes
- Look at `lib/tasks` in the Strata SDK repository for examples of custom Rake tasks

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

baonguyenNava:
## Ticket

Resolves #56 

## Changes

Adds instructions for running rails runner in our deployed environments as well locally.

## Context for reviewers

A developer will need to be able to update a user's attributes for testing with attribute based access control.

## Testing

> Provide evidence that the code works as expected. Explain what was done for testing and the results of the test plan. Include screenshots, [GIF demos](https://www.cockos.com/licecap/), shell commands or output to help show the changes working as expected. ProTip: you can drag and drop or paste images into this textbox.

♻️ Environment destroyed ♻️


reporting-app/lib/tasks/users.rake
lorenyu: Reading this makes me think if it would be simpler to just run `rails runner 'User.find_by(email:"foo@navapbc.com").update(full_name:"Casey Jones",region:"Northwest")'` and update the instruction doc for how to run that locally and in the environment via bin/run-command.

lorenyu (commented):
Left a thought about whether we could just use rails runner + some documentation instead of a Rake task

Issue #72, opened 2025-12-04
Foundation: Create HoursDataEntry Model and Table


nkutub:
## User Story

As a **system**  
I want **raw hours data to be stored in a dedicated table**  
So that **we can track all hours submissions from external sources with full audit trail**

---

## Acceptance Criteria

- [ ] `hours_data_entries` table exists in database
- [ ] Table has all required columns with correct data types
- [ ] All indexes are created (certification_id, source_type+source_id, period dates, reported_at)
- [ ] `HoursDataEntry` model exists with all validations
- [ ] Model includes required scopes (for_certification, by_category, from_api, from_batch, in_period)
- [ ] Custom validations work (period_end_after_start, certification_exists, period_within_lookback)
- [ ] Lookup logic finds latest certification by member_id only (no certification_date needed)
- [ ] Factory exists for testing
- [ ] Full RSpec test coverage (validations, scopes, business rules)

---

### Database Migration

Create migration:

```bash
rails g migration CreateHoursDataEntries
```

**Table Structure:**

```ruby
create_table :hours_data_entries, id: :uuid do |t|
  t.uuid :certification_id, null: false
  t.string :category, null: false
  t.decimal :hours, precision: 8, scale: 2, null: false
  t.date :period_start, null: false
  t.date :period_end, null: false
  t.string :source_type, null: false
  t.uuid :source_id
  t.datetime :reported_at, null: false
  t.jsonb :metadata, default: {}
  t.timestamps
end

add_index :hours_data_entries, :certification_id
add_index :hours_data_entries, [:source_type, :source_id]
add_index :hours_data_entries, [:period_start, :period_end]
add_index :hours_data_entries, :reported_at
```

### Model File

**Location:** `app/models/hours_data_entry.rb`

**Key Constants:**

- `ACTIVITY_TYPES = %w[employment community_service education]`
- `SOURCE_TYPE_API = "api"`
- `SOURCE_TYPE_BATCH = "batch_upload"`

**Validations:**

- Standard Rails validations for presence, inclusion, numericality
- Custom validation methods for business rules

**Scopes:**

- `for_certification(cert_id)`
- `by_category(category)`
- `from_api`
- `from_batch(batch_id)`
- `in_period(start_date, end_date)`

### Test File

**Location:** `spec/models/hours_data_entry_spec.rb`

**Test Coverage:**

- All validation rules
- All scopes
- Custom validation methods
- Edge cases (boundary values)

### Factory

[...]

PR #91, opened 2025-12-09
Create ExParteActivity Model and Table

nkutub:
## Ticket

Resolves [HOURS-001: Create ExParteActivity Model and Table](https://github.com/navapbc/oscer/issues/72)

## Changes

- **Added** migration `20251210000001_create_ex_parte_activities.rb` creating the `ex_parte_activities` table with:
  - UUID primary key
  - `member_id` (required) - enables hours-first save order
  - `certification_id` (nullable) - for pending entries before cert creation
  - `category`, `hours`, `period_start`, `period_end` - activity data
  - `outside_period` flag - for activities outside lookback period
  - `source_type`, `source_id` - audit trail
  - `reported_at`, `metadata` - submission context
  - Indexes including partial index for pending entries
  - PostgreSQL column comments for schema documentation

- **Added** `ExParteActivity` model with:
  - Validations for all required fields
  - 4 scopes: `for_certification`, `pending_for_member`, `by_category`, `in_period`
  - 2 methods: `pending?`, `link_to_certification!`

- **Added** factory `ex_parte_activities_factory.rb` with traits: `:pending`, `:employment`, `:community_service`, `:education`

- **Added** RSpec tests (29 examples) covering validations, scopes, and methods

## Context for reviewers

This is the foundation model for the Hours Automation epic. It stores hours data submitted via API or batch upload ("ex parte" sources) as opposed to manually reported hours.

**Key design decisions:**
- `member_id` is always required, `certification_id` is nullable to support "hours-first save order" pattern (hours saved before certification creation)
- `outside_period` flag will be set by the service layer (HOURS-004), not by model callbacks, following `.cursorrules` guidance to avoid cross-aggregate queries in models
- Model kept intentionally lean - presentation logic and convenience scopes deferred until needed

**Related tickets:**
- HOURS-004 (ExParteActivityService) will handle creating entries and setting `outside_period`
- HOURS-005 (HoursComplianceDeterminationService) will use `for_certification` scope for aggregation

## Testing

```bash
# Run migration
bin/rails db:migrate

# Run model tests
bundle exec rspec spec/models/ex_parte_activity_spec.rb

```

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/ex_parte_activity.rb
+++ b/reporting-app/app/models/ex_parte_activity.rb
@@ -6,8 +6,8 @@
 # reported hours from ActivityReportApplicationForm. Ex parte hours are
 # auto-verified and don't require staff review.
 #
-# @note The `outside_period` flag should be set by the service layer
-#   (ExParteActivityService) when creating entries, not by callbacks.
+# Activities are linked to certifications through member_id - since there's
+# only one active certification per member at a time, the relationship is implicit.
 #
 class ExParteActivity < ApplicationRecord
   include Strata::Attributes
@@ -40,16 +40,5 @@ class ExParteActivity < ApplicationRecord
 
   # --- Scopes ---
 
-  scope :for_certification, ->(cert_id) { where(certification_id: cert_id) }
-  scope :pending_for_member, ->(member_id) { where(member_id: member_id, certification_id: nil) }
-
-  # --- Instance Methods ---
-
-  def pending?
-    certification_id.nil?
-  end
-
-  def link_to_certification!(cert_id)
-    update!(certification_id: cert_id)
-  end
+  scope :for_member, ->(member_id) { where(member_id: member_id) }
 end
--- a/reporting-app/db/migrate/20251210000001_create_ex_parte_activities.rb
+++ b/reporting-app/db/migrate/20251210000001_create_ex_parte_activities.rb
@@ -4,19 +4,15 @@ class CreateExParteActivities < ActiveRecord::Migration[7.2]
   def change
     create_table :ex_parte_activities, id: :uuid, comment: "Hours data from external sources (API/batch) for compliance calculation" do |t|
       t.string :member_id, null: false,
-               comment: "Member reference - always required (allows hours-first save order)"
-      t.uuid :certification_id,
-             comment: "Certification reference - nullable for pending entries before cert creation"
+               comment: "Member reference - always required"
       t.string :category, null: false,
                comment: "Activity category: employment, community_service, education"
       t.decimal :hours, precision: 8, scale: 2, null: false,
-                comment: "Hours worked/volunteered (max 744 = 31 days × 24 hours)"
+                comment: "Hours worked/volunteered (max 8760 = 365 days × 24 hours)"
       t.date :period_start, null: false,
              comment: "Activity period start date"
       t.date :period_end, null: false,
              comment: "Activity period end date"
-      t.boolean :outside_period, default: false, null: false,
-                comment: "True if activity dates are outside certification lookback period"
       t.string :source_type, null: false,
                comment: "Source type: 'api' or 'batch_upload'"
       t.string :source_id,
@@ -26,23 +22,12 @@ class CreateExParteActivities < ActiveRecord::Migration[7.2]
 
     add_index :ex_parte_activities, :member_id,
               name: "index_ex_parte_activities_on_member_id",
-              comment: "Lookup entries by member (for pending entry lookups)"
-    add_index :ex_parte_activities, :certification_id,
-              name: "index_ex_parte_activities_on_certification_id",
-              comment: "Lookup entries by certification"
+              comment: "Lookup entries by member"
     add_index :ex_parte_activities, [ :source_type, :source_id ],
               name: "index_ex_parte_activities_on_source",
               comment: "Source tracking (batch upload lookups)"
     add_index :ex_parte_activities, [ :period_start, :period_end ],
               name: "index_ex_parte_activities_on_period",
               comment: "Date range queries"
-    add_index :ex_parte_activities, [ :member_id, :certification_id ],
-              where: "certification_id IS NULL",
-              name: "idx_ex_parte_activities_pending",
-              comment: "Partial index for pending entries (certification_id IS NULL)"
-
-    add_foreign_key :ex_parte_activities, :certifications,
-                    column: :certification_id,
-                    on_delete: :nullify
   end
 end
--- a/reporting-app/spec/factories/ex_parte_activities_factory.rb
+++ b/reporting-app/spec/factories/ex_parte_activities_factory.rb
@@ -3,29 +3,28 @@
 FactoryBot.define do
   factory :ex_parte_activity do
     member_id { Faker::NationalHealthService.british_number }
-    certification_id { create(:certification).id }
     category { ExParteActivity::ALLOWED_CATEGORIES.sample }
     hours { Faker::Number.between(from: 1.0, to: 200.0).round(2) }
     period_start { Date.current.beginning_of_month }
     period_end { Date.current.end_of_month }
-    outside_period { false }
     source_type { ExParteActivity::SOURCE_TYPES[:api] }
     source_id { nil }
 
-    trait :pending do
-      certification_id { nil }
-    end
-
     trait :employment do
-      category { "employment" }
+      category { 'employment' }
     end
 
     trait :community_service do
-      category { "community_service" }
+      category { 'community_service' }
     end
 
     trait :education do
-      category { "education" }
+      category { 'education' }
+    end
+
+    trait :from_batch do
+      source_type { ExParteActivity::SOURCE_TYPES[:batch] }
+      source_id { SecureRandom.uuid }
     end
   end
 end
--- a/reporting-app/db/schema.rb
+++ b/reporting-app/db/schema.rb
@@ -115,20 +115,16 @@ ActiveRecord::Schema[7.2].define(version: 2025_12_10_000001) do
   end
 
   create_table "ex_parte_activities", id: :uuid, default: -> { "gen_random_uuid()" }, comment: "Hours data from external sources (API/batch) for compliance calculation", force: :cascade do |t|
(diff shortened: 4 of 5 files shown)


reporting-app/app/models/ex_parte_activity.rb
```diff
@@ -0,0 +1,62 @@
...
+  MAX_HOURS = 744
+
+  # --- Associations ---
+
+  belongs_to :certification, optional: true
+
+  # --- Validations ---
+
+  validates :member_id, presence: true
+  validates :category, presence: true, inclusion: { in: ALLOWED_CATEGORIES }
```
lorenyu: why not just use an `enum` type

reporting-app/app/models/ex_parte_activity.rb
```diff
@@ -0,0 +1,62 @@
...
+  validates :hours, presence: true,
+                    numericality: { greater_than: 0, less_than_or_equal_to: MAX_HOURS }
+  validates :period_start, presence: true
+  validates :period_end, presence: true
```
lorenyu: I think you can use SDK's DateRange type

reporting-app/app/models/ex_parte_activity.rb
```diff
@@ -0,0 +1,62 @@
...
+  validates :source_type, presence: true, inclusion: { in: ALLOWED_SOURCE_TYPES }
```
lorenyu: enum type?

reporting-app/app/models/ex_parte_activity.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  SOURCE_TYPE_API = "api"
+  SOURCE_TYPE_BATCH = "batch_upload"
+  ALLOWED_SOURCE_TYPES = [ SOURCE_TYPE_API, SOURCE_TYPE_BATCH ].freeze
+
+  MAX_HOURS = 744
+
+  # --- Associations ---
+
+  belongs_to :certification, optional: true
```
lorenyu: Follow [the data modeling guidelines](https://github.com/navapbc/strata-sdk-rails/blob/68483d71a15444796864ae50ff7073812c336247/docs/contributing/data-modeling-guidelines.md) and avoid adding associations between aggregates

reporting-app/spec/factories/ex_parte_activities_factory.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :ex_parte_activity do
+    member_id { Faker::NationalHealthService.british_number }
+    certification
+    category { ExParteActivity::ALLOWED_CATEGORIES.sample }
+    hours { Faker::Number.between(from: 1.0, to: 80.0).round(2) }
```
lorenyu: should be able to go above 80

reporting-app/db/migrate/20251210000001_create_ex_parte_activities.rb
```diff
@@ -0,0 +1,58 @@
...
+             comment: "Activity period start date"
+      t.date :period_end, null: false,
+             comment: "Activity period end date"
+      t.boolean :outside_period, default: false, null: false,
+                comment: "True if activity dates are outside certification lookback period"
+      t.string :source_type, null: false,
+               comment: "Source type: 'api' or 'batch_upload'"
+      t.uuid :source_id,
+             comment: "Source record ID (e.g., HoursDataBatchUpload ID for batch_upload)"
+      t.datetime :reported_at, null: false,
```
lorenyu: question: is this like when the data was originally obtained by the state in their system?

reporting-app/db/migrate/20251210000001_create_ex_parte_activities.rb
```diff
@@ -0,0 +1,58 @@
...
+                 comment: "When the data was submitted"
+      t.jsonb :metadata, default: {},
```
lorenyu: do we have any known expected uses for this at the moment?

reporting-app/db/migrate/20251210000001_create_ex_parte_activities.rb
```diff
@@ -0,0 +1,58 @@
...
+      t.string :category, null: false,
+               comment: "Activity category: employment, community_service, education"
+      t.decimal :hours, precision: 8, scale: 2, null: false,
+                comment: "Hours worked/volunteered (max 744 = 31 days × 24 hours)"
+      t.date :period_start, null: false,
+             comment: "Activity period start date"
+      t.date :period_end, null: false,
+             comment: "Activity period end date"
+      t.boolean :outside_period, default: false, null: false,
+                comment: "True if activity dates are outside certification lookback period"
```
lorenyu: don't think we should have this, we should just tie the data to the member not the certification

reporting-app/app/models/ex_parte_activity.rb
```diff
@@ -0,0 +1,62 @@
...
+#
+# @note The `outside_period` flag should be set by the service layer
+#   (ExParteActivityService) when creating entries, not by callbacks.
+#
+class ExParteActivity < ApplicationRecord
+  ALLOWED_CATEGORIES = %w[employment community_service education].freeze
+
+  SOURCE_TYPE_API = "api"
+  SOURCE_TYPE_BATCH = "batch_upload"
+  ALLOWED_SOURCE_TYPES = [ SOURCE_TYPE_API, SOURCE_TYPE_BATCH ].freeze
```
lorenyu: minor nit can we group these to DRY it up a bit
```suggestion
  SOURCE_TYPE = {
    API: "api",
    SOURCE_TYPE_BATCH: "batch_upload"
  }
  ALLOWED_SOURCE_TYPES = SOURCE_TYPE.values.freeze
```

reporting-app/app/models/ex_parte_activity.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  MAX_HOURS = 744
```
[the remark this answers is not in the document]
nkutub: chnaged to MAX_HOURS_PER_YEAR

reporting-app/db/migrate/20251210000001_create_ex_parte_activities.rb
nkutub: see comment above about it being tied to memeber, but with our app more certification centric UI , we should maintain data about how it related to certs even if record officially is related to member. otherwise we are doing on the fly calcs/queries that slow things down and our data is inferred , not direct.

reporting-app/db/migrate/20251210000001_create_ex_parte_activities.rb
nkutub: yes this was intended for that , again I have seen this needed in my experience specially when troubleshooting and reporting.

nkutub: but I will remove for now since it speculative at this time

reporting-app/app/models/ex_parte_activity.rb
nkutub: I thought of that but since we are likely not using this table on a form and don't need complex valiudation it might be overkill here our three line validation does all we need :
`validates :period_start, presence: true
   validates :period_end, presence: true
   validate :period_end_after_start`

reporting-app/app/models/ex_parte_activity.rb
lorenyu: My understanding is Rails allows enums to be backed by strings (vs integers)

nkutub: yes but we have a bigger issue is the enums are model specific , here we need both models to have the same categories, I decided to use best practice and extract the shared constant to a module shared in Activity and ExParteActivity

reporting-app/app/models/ex_parte_activity.rb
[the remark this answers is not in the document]
nkutub: the schema file is also very easy to access

reporting-app/app/models/ex_parte_activity.rb
nkutub: I will change for future needs.

reporting-app/db/migrate/20251210000001_create_ex_parte_activities.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+class CreateExParteActivities < ActiveRecord::Migration[7.2]
+  def change
+    create_table :ex_parte_activities, id: :uuid, comment: "Hours data from external sources (API/batch) for compliance calculation" do |t|
+      t.string :member_id, null: false,
+               comment: "Member reference - always required (allows hours-first save order)"
+      t.uuid :certification_id,
```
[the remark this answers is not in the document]
lorenyu: i'm open to discussing but it adds complexity so feels like a potentially premature optimization. in particular, we now have to keep this field in sync whenever we add new certifications or add new ex parte activities which is two additional places we have to do DB writes that can go out of sync.

i don't think the query becomes much more complicated or less performant without this field. instead of

```
SELECT * FROM ex_parte_activities
WHERE ex_parte_activities.certification_id = :certification_id
``` 

it becomes 

```
SELECT * FROM ex_parte_activities
WHERE ex_parte_activities.member_id = :member_id
  AND ex_parte_activities.period_start < :certification_period_end
  AND ex_parte_activities.period_start > :certification_period_start
```

with an index on [member_id, period_start, period_start] we're fine, and we'll need an index on those columns anyways to support the data sync

reporting-app/db/migrate/20251210000001_create_ex_parte_activities.rb
lorenyu: got it thanks for explaining. yeah i can see that this is would be valuable.

reporting-app/db/migrate/20251210000001_create_ex_parte_activities.rb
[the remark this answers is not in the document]
lorenyu: ok sounds good. i think you have a point about simplifying reporting and i have a question about auditability, so maybe we will need this eventually, but if you're ok with adding it later when we need it then i'd prefer that route, appreciate your patience with me on this

PR #94, opened 2025-12-11
Staff may only access tasks for their region

### LINKED ISSUE #60: Authz: Staff May Only Access Tasks For Their Region
## User Story

As an OSCER staff member

I want to only be shown tasks that I am allowed to work

So that I can do my job successfully

## Acceptance Criteria
- Staff members should only see Tasks that they have access to
    - On the task list page, the task list should only show tasks that the staff member has access to
    - Staff members should only be able to open individual tasks that they have access to
- Tests should be written to verify access
    - See [Pundit documentation](https://github.com/varvet/pundit?tab=readme-ov-file#policy-specs) for examples of writing permissions tests

## Technical Notes
- This should most likely be accomplished by using [Pundit Scopes](https://github.com/varvet/pundit?tab=readme-ov-file#scopes)
- See [the RBAC tech spec](https://docs.google.com/document/d/1xKYfUCxyr2MQHlZ3RWX1TrI24J_HSnx2aGVCzyI59bI/edit?tab=t.0#heading=h.2gubjermh1fh), Example Policy Definition section, for a somewhat similar-ish example of what this could look like

MichaelCrawfordNava:
## Ticket

Resolves #60 

## Changes

- Added task policy

## Context for reviewers

Some of the changes aren't ideal currently and would be addressed if/when we move logic from OSCER into the SDK.

Longer term, I would hope that we would move the `TaskPolicy` into Strata. However, in order to do this, we also need to move the concept of a `User` into Strata (with the user's role and potentially region).

### Tasks Index (Task list page)

- In your local database, set your certifications to have different regions (for example, set 1 certification to have "Northwest", another to have "Southeast", etc.)
    - Do this in the certification's `certification_requirements` field
- Set your user to have a region that matches one or more of your local certifications
- Ensure your user has the "caseworker" or "admin" role
- Navigate to the staff tasks index page
- Verify that the staff index tasks page loaded and that you weren't redirected
- Verify that the only tasks that show on this page are tasks for certifications within your user's region
- Set your local user to have no role (or any role that isn't "caseworker" or "admin")
- Verify that you cannot load the staff index tasks page

### Tasks Show (Individual task page)

- Ensure that you have the "caseworker" or "admin" role
- Open an individual task that shares the same region as your user, and verify that it loads correctly
- Open an individual task that does **not** share the same region as your user, and verify that you cannot load that task
- Set your local user to have no role (or any role that isn't "caseworker" or "admin")
- Open an individual task and verify that you cannot load that task

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/certification.rb
+++ b/reporting-app/app/models/certification.rb
@@ -17,6 +17,7 @@ class Certification < ApplicationRecord
   validates :certification_requirements, presence: true
 
   scope :by_member_id, ->(member_id) { where(member_id:) }
+  scope :by_region, ->(region) { where("certification_requirements->>'region' = ?", region) }
 
   after_create_commit do
     Strata::EventManager.publish("CertificationCreated", { certification_id: id })
--- a/reporting-app/app/policies/task_policy.rb
+++ b/reporting-app/app/policies/task_policy.rb
@@ -33,11 +33,7 @@ class TaskPolicy < Strata::TaskPolicy
     def resolve
       return scope.none unless user&.staff?
 
-      # TODO: Once authorization is moved to Strata, move these joins/where to Strata::Task which can't be done in OSCER
-      scope
-      .joins("INNER JOIN certification_cases ON certification_cases.id = strata_tasks.case_id")
-      .joins("INNER JOIN certifications ON certifications.id = certification_cases.certification_id")
-      .where("certifications.certification_requirements->>'region' = ?", user.region)
+      scope.by_region(user.region)
     end
   end
 
--- a/reporting-app/config/initializers/strata_task_extentions.rb
+++ b/reporting-app/config/initializers/strata_task_extentions.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+Rails.application.config.to_prepare do
+  Strata::Task.class_eval do
+    scope :by_region, ->(region) {
+      tasks = arel_table
+      cases = CertificationCase.arel_table
+      certs = Certification.arel_table
+
+      join_conditions = tasks[:case_id].eq(cases[:id])
+        .and(tasks[:case_type].eq("CertificationCase"))
+
+      joins(
+        tasks.join(cases, Arel::Nodes::InnerJoin).on(join_conditions)
+          .join(certs, Arel::Nodes::InnerJoin).on(cases[:certification_id].eq(certs[:id]))
+          .join_sources
+      ).merge(Certification.by_region(region))
+    }
+  end
+end


reporting-app/app/models/certification.rb
```diff
@@ -68,4 +68,8 @@ def origin
   def member_name
     self&.member_data&.name
   end
+
+  def region
+    self&.certification_requirements&.region
```
giverm: I believe `self` isn't necessary on instance methods. 

```suggestion
  certification_requirements&.region
```

reporting-app/app/policies/task_policy.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+class TaskPolicy < ApplicationPolicy
```
giverm: There's a `Strata::Task` policy that was created last week as part of the Staff policy work. 
https://github.com/navapbc/oscer/blob/main/reporting-app/app/policies/strata/task_policy.rb

Would using that simplify the scope query? Either way, this should be consolidated

reporting-app/app/models/oscer_task.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class OscerTask < Strata::Task
```
giverm: I think I have a slight preference for extending the `Strata::Task` class like we had discussed before (https://github.com/navapbc/oscer/compare/main...giverm/test-scope), rather than creating an additional class between `Strata::Task` and `ReviewActivityReportTask`/`ReviewExemptionTask`

I'm open to either though. Not sure what others think

reporting-app/db/seeds/development.rb
```diff
@@ -1,12 +1,17 @@
 # frozen_string_literal: true
 
+# Disable email delivery during seeding to prevent letter_opener from opening browser tabs
+original_delivery_method = ActionMailer::Base.delivery_method
+ActionMailer::Base.delivery_method = :test
+
 # Create sample batch uploads for testing
-user = User.first || User.create!(email: "staff@example.com", uid: SecureRandom.uuid, provider: "login.gov")
+user = User.first || User.create!(email: "staff@example.com", uid: SecureRandom.uuid, provider: "login.gov", region: "Southeast", role: "admin")
```
giverm: The user factory has an `:as_admin` trait now

reporting-app/app/models/certification.rb
MichaelCrawfordNava: Yeah, I was just following the convention already set in the file. It seems like other instance methods are using `self`.

reporting-app/db/seeds/development.rb
MichaelCrawfordNava: I switched to using `FactoryBot` to take advantage of the `as_admin` trait

Issue #75, opened 2025-12-04
Foundation: Create ExParteActivityService


nkutub:
## User Story

As a **member**  
I want **my hours to be automatically calculated against the 80-hour target**  
So that **I know if I'm compliant without waiting for manual review**

---

## Acceptance Criteria

- [x] `HoursComplianceDeterminationService` class exists
- [x] **PRIMARY:** `.determine(kase)` method for business process integration at EX_PARTE_CE_CHECK step
- [x] **FALLBACK:** `.calculate(certification_id)` method for async recalculation of existing certs
- [x] Both methods aggregate hours from BOTH sources (ExParteActivity + Activity)
- [x] Includes ex parte hours (ExParteActivity - always counted)
- [x] Includes manual hours (Activity - only if activity_report_approval_status == "approved")
- [x] TARGET_HOURS configurable via environment variable (default: 80)
- [x] Outcome determined correctly (>= 80 = compliant, < 80 = not_compliant)
- [x] Correct reason codes used based on outcome (hours_reported_compliant, hours_insufficient)
- [x] determination_data includes all required fields
- [x] `calculation_method` field distinguishes "business_process" vs "async_recalculation"
- [x] Service creates automated determination via `record_determination!`
- [x] Full RSpec test coverage (18 examples)
- [x] Business process updated to call service

---

### Service File

**Location:** `app/services/hours_compliance_determination_service.rb`

**Implemented Methods:**

- `determine(kase)` - Called by business process at EX_PARTE_CE_CHECK step
- `calculate(certification_id)` - Called by CalculateComplianceJob for existing certs
- `aggregate_hours_for_certification(certification)` - Combines ex parte and manual hours
- `aggregate_ex_parte_hours(certification_id)` - Sums ExParteActivity hours
- `aggregate_manual_hours(certification)` - Sums approved Activity hours

### CertificationCase Method

**Added:** `determine_ce_hours_compliance(outcome, hours_data)`

- Creates determination with hours-based data
- Sets `activity_report_approval_status` to "approved" on compliance
- Publishes `DeterminedRequirementsMet` or `DeterminedRequirementsNotMet` events
- Follows pattern of existing `determine_ex_parte_exemption`

### Business Process Integration

**File:** `app/business_processes/certification_business_process.rb`

```ruby
system_process(EX_PARTE_COMMUNITY_ENGAGEMENT_CHECK_STEP, ->(kase) {
  HoursComplianceDeterminationService.determine(kase)
})
```

[...]

PR #96, opened 2025-12-12
Create hours compliance determination service

nkutub:
## Ticket

Resolves  [HoursComplianceDeterminationService](https://github.com/navapbc/oscer/issues/75)

### HoursComplianceDeterminationService

**Key changes:**
1. Now queries ExParteActivity by `member_id` instead of `certification_id`
2. **Added lookback period filtering** - only counts hours within the certification's `continuous_lookback_period`
3. Renamed `manual_hours` → `member_hours` for clarity (internal variable/method names)

**New filtering logic:**
```ruby
def aggregate_ex_parte_hours(member_id, lookback_period)
  entries = ExParteActivity.for_member(member_id)
  
  if lookback_period.present?
    entries = entries.where(
      "period_start >= ? AND period_end <= ?",
      lookback_period.start.to_date,
      lookback_period.end.to_date.end_of_month
    )
  end
  # ...
end
```

### Other Fixes

- Fixed `member_status` visibility in `CertificationCase` (was accidentally private)

### Why remove `certification_id`?

The business rule is **one active certification per member at a time**. This means:
- Activities can be linked to certifications implicitly through `member_id`
- No need to explicitly track which certification an activity belongs to
- Simplifies the data model and eliminates nullable foreign keys

### Why add lookback period filtering?

Without `certification_id`, we needed a way to ensure we only count hours that fall within the certification's reporting period. The filtering uses:
- `certification.certification_requirements.continuous_lookback_period` 
- Compares against `ExParteActivity.period_start` and `period_end`

### Design Decisions

1. **No `belongs_to :certification`** - Follows `.cursorrules` pattern of not adding associations between aggregate roots
2. **Lookback filtering in service** - Keeps model simple, business logic in service layer
3. **Constants over enums** - `SOURCE_TYPES` uses simple hash constants per `.cursorrules`

### Run the specs

```bash
# ExParteActivity model tests
bundle exec rspec spec/models/ex_parte_activity_spec.rb

# HoursComplianceDeterminationService tests
bundle exec rspec spec/services/hours_compliance_determination_service_spec.rb

# CertificationBusinessProcess tests (includes hours compliance flow)
bundle exec rspec spec/business_processes/certification_business_process_spec.rb

# All related specs
bundle exec rspec spec/models/ex_parte_activity_spec.rb \
                  spec/services/hours_compliance_determination_service_spec.rb \
                  spec/business_processes/certification_business_process_spec.rb
```

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/spec/models/determination_spec.rb
+++ b/reporting-app/spec/models/determination_spec.rb
@@ -3,6 +3,14 @@
 require 'rails_helper'
 
 RSpec.describe Determination, type: :model do
+  before do
+    # Prevent auto-triggering business process during test setup
+    allow(Strata::EventManager).to receive(:publish).and_call_original
+    allow(HoursComplianceDeterminationService).to receive(:determine)
+    allow(ExemptionDeterminationService).to receive(:determine)
+    allow(NotificationService).to receive(:send_email_notification)
+  end
+
   describe 'enums' do
     describe 'decision_method' do
       it 'defines the decision_method enum with correct values' do
--- a/reporting-app/spec/requests/staff/certification_batch_uploads_spec.rb
+++ b/reporting-app/spec/requests/staff/certification_batch_uploads_spec.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require 'support/query_count_matchers'
 
 RSpec.describe "Staff::CertificationBatchUploads", type: :request do
   include Warden::Test::Helpers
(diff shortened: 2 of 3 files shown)


reporting-app/app/models/certification_case.rb
```diff
@@ -105,10 +105,47 @@ def determine_ex_parte_exemption(eligibility_fact)
...
+
+  # Called by HoursComplianceDeterminationService for hours compliance determination
+  # @param outcome [Symbol] :compliant or :not_compliant
+  # @param hours_data [Hash] aggregated hours data
+  # @param trigger_workflow [Boolean] whether to trigger events/notifications (default: true)
+  #   - true: Called from business process, triggers state changes, events, notifications
+  #   - false: Called from async recalculation, only records determination
+  def determine_ce_hours_compliance(outcome, hours_data, trigger_workflow: true)
+    certification = Certification.find(self.certification_id)
+    calculation_method = trigger_workflow ? "business_process" : "async_recalculation"
```
giverm: Can add constants instead of using magic strings here?

giverm (commented):
Looks good. Just a few minor comments

Issue #139, opened 2025-12-19
Clean up and document staff authorization patterns


giverm:
## Background
The staff authorization implementation has evolved organically, leading to inconsistencies and unclear patterns. This issue tracks work to clean up authorization policies, fix security gaps, and establish clear documentation for the two-layer authorization model.

## Authorization Architecture
**Two-layer model:**
1. **Page access** - StaffPolicy checks if user is staff (authorize :staff)
2. **Data scoping** - Resource policies filter data by region (policy_scope)
3. **Record access** - Resource policies check region for specific records (authorize @record)

## Completed
- [x] CertificationCasesController - Remove unused methods, add documentation (#138)
- [x] InformationRequestsController - Add region-based authorization, fix security gap (#138)

## In Progress
- [ ] TasksController - Consolidate policies into Strata::TaskPolicy (branch: giverm/staff-policy-cleanup-tasks)
- [ ] Fix pick_up_next_task action to scope tasks by region

## Pending
- [ ] AdminController, Staff::CertificationBatchUploadsController, Staff::UsersController - Group cleanup
- [ ] MembersController - Needs business requirements clarification (can staff search across all regions?)
- [ ] CertificationsController - Needs API access discussion (shared with public API)

## Security Issues Found
- ✅ InformationRequests not scoped by region - Fixed in #138
- ⚠️ pick_up_next_task action does not scope by region - To be fixed

## Goals
1. **Consistency** - All controllers follow the same authorization pattern
2. **Security** - All data properly scoped by region
3. **Clarity** - Clear documentation on when to use which policy
4. **Maintainability** - Remove redundant/unused policy methods

## Related
- Architecture Decision Record on ABAC: docs/decisions/adr-authorization.md
- #37 Authorization implementation

## Non-Goals
This work focuses on cleaning up existing patterns, not redesigning the authorization architecture.

PR #150, opened 2025-12-19
Clean up TasksController authorization and fix pick_up_next_task region scoping

giverm:
Part of #139

This commit consolidates task authorization and ensures pick_up_next_task only assigns tasks from the user's region.

Changes:
- Consolidated all task authorization logic into Strata::TaskPolicy
- Removed redundant app/policies/task_policy.rb (logic moved to Strata::TaskPolicy)
- Updated OscerTask.policy_class to reference Strata::TaskPolicy
- Added region-based scoping to Strata::TaskPolicy::Scope
- Overrode pick_up_next_task to use policy_scope instead of Strata::Task.assign_next_task_to
- Simplified authorization by skipping StaffController's authorize_staff_access and using authorize_task_access before_action for collection actions
- Updated spec to reference Strata::TaskPolicy
- Added comprehensive tests for pick_up_next_task covering:
  * Assigning tasks from user's region
  * Not assigning tasks from other regions
  * Handling no available tasks scenario
  * Picking first task by due date when multiple available

The authorization model is now:
- Collection actions (index, pick_up_next_task): authorize Strata::Task class
- Instance actions (show, update, etc.): authorize @task instance in set_task
- All authorization uses Strata::TaskPolicy with region-based access control

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/controllers/tasks_controller.rb
+++ b/reporting-app/app/controllers/tasks_controller.rb
@@ -29,6 +29,7 @@ class TasksController < Strata::TasksController
   # which doesn't filter by region, so we override to inject policy_scope.
   def pick_up_next_task
     # Scope tasks to user's region before finding next unassigned task
+    # Default scope on Strata::Task is ordered by :due_on
     task = policy_scope(Strata::Task).incomplete.unassigned.first
 
     if task && task.assign(current_user.id)
--- a/reporting-app/app/policies/strata/task_policy.rb
+++ b/reporting-app/app/policies/strata/task_policy.rb
@@ -53,9 +53,5 @@ module Strata
     def in_region?
       user.region == ::TaskService.get_region_for_task(record)
     end
-
-    def staff_in_region?
-      staff? && in_region?
-    end
   end
 end


reporting-app/app/controllers/tasks_controller.rb
```diff
@@ -22,6 +24,22 @@ def assign
...
+      flash["task-message"] = I18n.t("strata.tasks.messages.task_picked_up")
+      redirect_to task_path(task)
+    else
+      flash["task-message"] = I18n.t("strata.tasks.messages.no_tasks_available")
+      redirect_to tasks_path
+    end
+  end
+
   def request_information
     set_task
```
nkutub: per our convention should be in before action . we set instance vars in callbacks

reporting-app/app/policies/strata/task_policy.rb
```diff
@@ -2,27 +2,60 @@
...
   class TaskPolicy < ::StaffPolicy
-    # Custom actions for TasksController
-    def assign?
+    # Collection/page-level actions - any staff can access
+    def index?
       staff?
     end
 
-    def request_information?
+    def pick_up_next_task?
```
nkutub: this was confusing at first can we add the following method comment header:
```
# Collection/page-level actions - any staff can access
# (actual task filtering is done via policy_scope in the controller)
```

reporting-app/app/controllers/tasks_controller.rb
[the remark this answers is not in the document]
baonguyenNava: I attempted to change this but it exposes a brittle inheritance with Strata::TaskControllers. I was unable to get the tests to pass with redefining the before_action hook. I'll have to look into how we can do this better.

Issue #191, opened 2026-01-14
Story 5: Environment Configuration


nkutub:
**User Story:**
As a **DevOps engineer**, I want the API secret key managed via environment variables so that secrets are securely configured across all environments.

**Files:**

- `infra/reporting-app/app-config/env-config/environment_variables.tf` — add `API_SECRET_KEY` to secrets
- `reporting-app/local.env.example` — document `API_SECRET_KEY=`

**Acceptance Criteria:**

- [ ] `API_SECRET_KEY` added to infra secrets config (auto-generated via SSM)
- [ ] `API_SECRET_KEY` documented in `local.env.example`
- [ ] Key is at least 32 bytes (256 bits)
- [ ] App fails fast with helpful error if `API_SECRET_KEY` missing

**Infrastructure Configuration:**

Add to `infra/reporting-app/app-config/env-config/environment_variables.tf`:

```hcl
locals {
  secrets = {
    SECRET_KEY_BASE = {
      manage_method     = "generated"
      secret_store_name = "/${var.app_name}-${var.environment}/service/rails-secret-key-base"
    }
    # Add this:
    API_SECRET_KEY = {
      manage_method     = "generated"
      secret_store_name = "/${var.app_name}-${var.environment}/service/api-secret-key"
    }
  }
}
```

This uses the existing `modules/secrets` pattern which:

- Auto-generates a 64-character secure random value
- Stores it in AWS SSM Parameter Store
- Injects it as an environment variable into the service

**Local Development:**

Add to `reporting-app/local.env.example`:

```bash
# API Authentication (generate with: openssl rand -base64 32)
API_SECRET_KEY=
```

**Verification:**

- [ ] `terraform plan` shows new SSM parameter
- [ ] Deployed service has `API_SECRET_KEY` env var
- [ ] Local development works with `.env` key
- [ ] Story 3 tests pass in CI

**Note:** No code tests required — this is configuration only. Verification is via successful Story 3 tests.

---

Issue #190, opened 2026-01-14
Story 4: Implement Real `state_system?` in Policies


nkutub:
**User Story:**
As a **developer**, I want policies to correctly identify API clients as state systems so that authorization rules apply consistently across UI and API access.

**Files:**

- `app/policies/certification_policy.rb` (modify existing)
- `spec/policies/certification_policy_spec.rb` (modify existing)

**Current state (stubbed):**

```ruby
def state_system?
  # TODO: once we have user types
  true
end
```

**Change to:**

- [ ] Check if user responds to `state_system?` and returns true
- [ ] Keep existing policy methods unchanged (they already use `state_system?`)

**Test Coverage:**

- [ ] `state_system?` returns `true` for `ApiClient`
- [ ] `state_system?` returns `false` for regular `User`
- [ ] Policy methods (`create?`, `update?`) return `true` for `ApiClient`
- [ ] Policy methods return appropriate value for regular users

**Test example:**

```ruby
RSpec.describe CertificationPolicy do
  describe "#state_system?" do
    context "with ApiClient" do
      let(:user) { ApiClient.new }
      let(:policy) { described_class.new(user, build(:certification)) }

      it "returns true" do
        expect(policy).to be_state_system
      end

      it "allows create" do
        expect(policy.create?).to be true
      end
    end

    context "with regular User" do
      let(:user) { build(:user) }
      let(:policy) { described_class.new(user, build(:certification)) }

      it "returns false" do
        expect(policy).not_to be_state_system
      end
    end
  end
end
```

**Implementation:**

```ruby
def state_system?
  user.respond_to?(:state_system?) && user.state_system?
end
```

**Note:** Lines 13, 17, 21, 25 already call `state_system?` — no changes needed there.

---

Issue #189, opened 2026-01-14
Story 3: Add Authentication to ApiController


nkutub:
**User Story:**
As a **state system integrator**, I want the API to require valid HMAC signatures so that unauthorized clients cannot submit certification data.

**Files:**

- `app/controllers/api_controller.rb` (modify existing)
- `spec/requests/api/hours_spec.rb` (modify existing or create)
- `spec/requests/api/certifications_spec.rb` (update existing — add auth headers)

**Existing code to keep:**

- `include Pundit::Authorization` ✅
- `render_errors` method ✅

**⚠️ Breaking Change:** Existing API tests use `let(:valid_headers) { {} }` — must update to include HMAC auth headers or all API tests will fail.

**Add:**

- [ ] `before_action :authenticate_api_request!`
- [ ] Sets `@current_api_client = ApiClient.new` on success
- [ ] Returns 401 using existing `render_errors` on failure (with `and return` to halt)
- [ ] `pundit_user` method returns `@current_api_client`
- [ ] Update existing `spec/requests/api/certifications_spec.rb` to use auth headers

**Test Coverage:**

- [ ] Request with valid HMAC signature returns 200
- [ ] Request with invalid signature returns 401 with error message
- [ ] Request with missing Authorization header returns 401
- [ ] Pundit authorization works with `ApiClient`

**Test example:**

```ruby
RSpec.describe "API Hours", type: :request do
  let(:secret) { ENV.fetch("API_SECRET_KEY") }
  let(:body) { { member_id: "123", hours: 80 }.to_json }
  let(:signature) { ApiAuthenticator.new(secret_key: secret).sign(body: body) }

  def api_headers(sig: signature)
    {
      "Authorization" => "HMAC sig=#{sig}",
      "Content-Type" => "application/json"
    }
  end

  describe "POST /api/hours" do
    context "with valid authentication" do
      it "returns success" do
        post "/api/hours", params: body, headers: api_headers
        expect(response).to have_http_status(:ok)
      end
    end

    context "with invalid signature" do
      it "returns 401" do
        post "/api/hours", params: body, headers: api_headers(sig: "invalid")
        expect(response).to have_http_status(:unauthorized)
        expect(json_response["errors"]).to include(/Signature verification failed/)
      end
    end

    context "without Authorization header" do
      it "returns 401" do
        post "/api/hours", params: body, headers: { "Content-Type" => "application/json" }
        expect(response).to have_http_status(:unauthorized)
[...]

PR #197, opened 2026-01-21
Add API authentication

### LINKED ISSUE #188: Story 2: Create ApiClient Model
**User Story:**
As a **developer**, I want an ApiClient model that represents authenticated API callers so that Pundit policies can authorize API requests using existing infrastructure.

**Files:**

- `app/models/api_client.rb` (new)
- `spec/models/api_client_spec.rb` (new)

**Acceptance Criteria:**

- [ ] Implements `state_system?` returning `true`
- [ ] Implements `staff?` returning `false`
- [ ] Implements `member?` returning `false`
- [ ] Satisfies `ApplicationPolicy` requirement for non-nil user

**Test Coverage:**

- [ ] `state_system?` returns `true`
- [ ] `staff?` returns `false`
- [ ] `member?` returns `false`

**Test example:**

```ruby
RSpec.describe ApiClient do
  subject(:api_client) { described_class.new }

  it { is_expected.to be_state_system }
  it { is_expected.not_to be_staff }
  it { is_expected.not_to be_member }
end
```

**Pseudocode:**

```ruby
class ApiClient
  # Represents an authenticated API client (state system)
  # Used by Pundit policies for authorization checks

  def state_system?
    true
  end

  def staff?
    false
  end

  def member?
    false
  end
end
```

**Notes:**

- Plain Ruby class, not ActiveRecord
- Required because `ApplicationPolicy#initialize` raises `Pundit::NotAuthorizedError` if user is nil
- Policies check `user.state_system?` to authorize API access

---

baonguyenNava:
## Ticket

Resolves #188 #189 #190 #191 

## Changes

Add API client object to be used as the Pundit user
Add Strata SDK api authenticator into base API controller
Update Certification policy to check pundit user/ remove hardcoded checks
Update API unit tests with API auth
Add api secret via Terraform

## Testing

I'll need to verify testing via making API calls after deployment to dev

♻️ Environment destroyed ♻️

### DIFF
--- a/infra/reporting-app/app-config/env-config/environment_variables.tf
+++ b/infra/reporting-app/app-config/env-config/environment_variables.tf
@@ -20,5 +20,9 @@ locals {
       manage_method     = "generated"
       secret_store_name = "/${var.app_name}-${var.environment}/service/rails-secret-key-base"
     }
+    API_SECRET_KEY = {
+      manage_method     = "generated"
+      secret_store_name = "/${var.app_name}-${var.environment}/service/api-secret-key"
+    }
   }
 }
--- a/reporting-app/app/controllers/api/healthcheck_controller.rb
+++ b/reporting-app/app/controllers/api/healthcheck_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Api::HealthcheckController < ApiController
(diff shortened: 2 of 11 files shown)


giverm (approved):
Aside from the issues with the existing certification policy spec, this looks good to me.

PR #283, opened 2026-02-18
Add custom:region attribute to Cognito User Pool and OSCER User provisioning

nkutub:
## Summary

Adds `custom:region` attribute to Cognito User Pool schema to support region-based access control for staff users.

**Changes:**
- Added `region` custom attribute to Cognito User Pool schema (accessed as `custom:region` in tokens)
- Added `custom:region` to client's `read_attributes` and `write_attributes`
- Added `custom:region` to the user provisioning logic to sync user region with IdP attribute

## Why

The application uses region-based task assignment and access control. Staff users need a `region` attribute stored in Cognito so:
1. Region is included in ID tokens after authentication
2. Application can sync region to user records during SSO login
3. Policy scopes can filter tasks/cases by user's region

♻️ Environment destroyed ♻️

### DIFF
--- a/infra/reporting-app/service/identity_provider.tf
+++ b/infra/reporting-app/service/identity_provider.tf
@@ -51,4 +51,8 @@ module "identity_provider_client" {
   name          = "${local.prefix}${local.identity_provider_config.identity_provider_name}"
 
   user_pool_id = local.identity_provider_user_pool_id
+
+  # Explicit dependency to ensure the User Pool schema (including custom:region)
+  # is fully created before the client tries to reference it in read/write attributes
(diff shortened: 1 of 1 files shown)


reporting-app/spec/services/staff_user_provisioner_spec.rb
```diff
@@ -109,6 +125,16 @@
           .from("caseworker")
           .to("admin")
       end
+
+      it "updates region when changed in IdP" do
```
baonguyenNava: I think we need a test to change an existing user's region to nil.

baonguyenNava (approved):

Issue #110, opened 2025-12-15
Epic: Batch Intake Improvements


sumiat:
## Background
In November 2025, we released functionality to support with the batch intake process. Our current batch processing is manual - an admin uploads a file through our system. However, we can improve the scale and optimize for performance. This effort builds on our batch intake process and optimizes it. 

## Scope
- Scope to be determined based on learnings from #102 

## Summary

Build a scalable batch upload system that processes large CSV files containing member certification data. Files can arrive via multiple sources (web UI, API, FTP, cloud storage events) and must be processed reliably with full auditability while supporting millions of records per file.

## Business Value

- **Efficiency**: Process hundreds of thousands of records in minutes, not hours
- **Reliability**: Chunk-based processing ensures partial failures don't lose all progress
- **Auditability**: Full tracking of uploads with error reports for failed records
- **Flexibility**: Accept files from multiple sources with identical business logic
- **Scalability**: Support growth in file size and concurrent uploads

## Solution Overview

```mermaid
flowchart LR
    subgraph Sources["Upload Sources"]
        UI[Staff UI]
        API[API]
        FTP[SFTP]
        Event[Storage Event]
    end

    subgraph OSCER["OSCER Application"]
        Orchestrator[Orchestrator]
        Storage[(Cloud Storage)]
        Workers[Background Workers]
        DB[(Database)]
    end

    UI & API -->|"Signed URL"| Storage
    FTP & Event --> Storage
    Storage -->|"Trigger"| Orchestrator
    Orchestrator --> Workers
    Workers -->|"Stream chunks"| Storage
    Workers -->|"Process"| DB
```

**How it works:**

1. Files uploaded directly to cloud storage (bypassing Rails for large files)
2. Upload confirmed via API, triggering background processing
3. Workers stream file in chunks (1,000 records each)
4. Each chunk processed in parallel for speed
5. Audit logs track progress; only failed records stored individually
6. Dashboard shows aggregated status

## Processing Flow

```mermaid
sequenceDiagram
    participant Client
    participant Rails
    participant Storage as Cloud Storage
    participant Worker
    participant DB

    Client->>Rails: GET /presigned_url
    Rails-->>Client: {url, storage_key}
    Client->>Storage: PUT file (direct)
    Client->>Rails: POST /batch_uploads (confirm)
    Rails->>DB: Create BatchUpload (pending)
    Rails->>Worker: Enqueue ProcessBatchUploadJob

    loop Each 1,000-record chunk
[...]

PR #288, opened 2026-02-20
Replace per-record inserts with bulk insert pipeline for chunk processing

giverm:
Chunk jobs previously processed records one at a time: for each of 1,000 rows, validate → EXISTS duplicate check → INSERT cert → INSERT origin → synchronous business process events. This meant 2,000+ individual INSERT statements and 1,000 individual EXISTS queries per chunk.

This replaces the per-record loop with a 3-phase bulk pipeline:

Phase 1 - Validate all records:
  Runs validation on all 1,000 records, separating valid from invalid.
  No database queries needed.

Phase 2 - Batch duplicate check:
  Single WHERE IN query replaces 1,000 individual EXISTS queries.
  Also detects within-batch duplicates (same member_id + case_number +
  certification_date appearing multiple times in one chunk).

Phase 3 - Bulk persist + events:
  insert_all! for certifications and origins in one transaction (2 bulk
  INSERTs vs 2,000 individual INSERTs), then publishes CertificationCreated
  events per record to trigger the business process chain.

Controlled profiling (Docker, fresh DB, 5k all-valid rows) shows:
  - Dedup: 33ms batch vs 3,311ms individual (100x faster)
  - Insert: 620ms bulk vs ~3,700ms individual (4.5x faster)
  - Overall: ~98s/chunk vs ~107s/chunk (~8% faster)

The modest overall gain is because the synchronous Strata business process chain dominates at ~97s/chunk in both approaches.

Changes:

ProcessCertificationBatchChunkJob:
  - Replace single record loop with 3-phase pipeline
  - Add validate_all_records, filter_duplicates, bulk_create_certifications

UnifiedRecordProcessor:
  - Add validate_record (non-raising validation for phase 1)
  - Add find_existing_duplicates (batch WHERE IN query)
  - Add compound_key (member_id|case_number|date composite key)
  - Add bulk_persist! (insert_all! with event publishing)
  - Existing process() path unchanged for non-batch callers

Certification model:
  - Extract publish_created_event as class method (callable without an instance, needed by bulk_persist! which uses insert_all!)

BatchUploadErrors:
  - Add WITHIN_BATCH duplicate error code

Part of #110

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/jobs/process_certification_batch_chunk_job.rb
+++ b/reporting-app/app/jobs/process_certification_batch_chunk_job.rb
@@ -61,33 +61,14 @@ class ProcessCertificationBatchChunkJob < ApplicationJob
     results = { succeeded: 0, failed: 0, errors: [] }
     context = { batch_upload_id: batch_upload.id }
 
-    records.each_with_index do |record, index|
-      row_number = calculate_row_number(chunk_number, index)
+    # Phase 1: Validate all records
+    valid_entries = validate_all_records(records, chunk_number, processor, results)
 
-      begin
-        processor.process(record, context: context)
-        results[:succeeded] += 1
-      rescue UnifiedRecordProcessor::ProcessingError => e
-        results[:failed] += 1
-        results[:errors] << {
-          row_number: row_number,
-          error_code: e.code,
-          error_message: e.message,
-          row_data: record
-        }
-      rescue StandardError => e
-        # Catch unexpected errors (shouldn't happen, but safety net)
-        Rails.logger.error("Unexpected error processing row #{row_number}: #{e.class} - #{e.message}")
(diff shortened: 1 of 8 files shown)


nkutub (approved):
LGTM

giverm:
Closing without merging. The PR's own benchmarks point at the real bottleneck: not the database inserts, but the business-process chain that runs after each record is created — ~97s of every ~107s per chunk. Bulk-inserting the records shaves about 8% off the total, which isn't enough to justify carrying two ways of persisting chunk records side-by-side forever: the existing per-record loop that runs Rails callbacks normally, and a new bulk-insert path that skips callbacks and adds its own dedup step.

If batch performance becomes a real complaint, the right fix is at the process-chain layer (event / process pipelining), not at the insert layer. This branch stays in git history if the bulk-insert ideas are worth revisiting then.

PR #308, opened 2026-03-04
Add HMAC-authenticated API endpoints for batch CSV uploads

### LINKED ISSUE #209: Story 10: Create API::BatchUploadsController
## User Story

As a **state system integrator**, I want an API endpoint for batch uploads so that I can automate data submission.

## Approach

Uses **Active Storage Direct Upload** with HMAC authentication (per PR #258 decision). Custom `DirectUploadsController` enables API clients to get presigned URLs using HMAC auth, following the same pattern as other API endpoints.

## Files

- `app/controllers/direct_uploads_controller.rb` (new - HMAC auth for Active Storage)
- `app/controllers/api/certification_batch_uploads_controller.rb` (new)
- `config/routes.rb` (modify - add API routes and override Active Storage route)
- `spec/requests/direct_uploads_spec.rb` (new)
- `spec/requests/api/certification_batch_uploads_spec.rb` (new)

### Feature Flag Gating
- [ ] All endpoints return 404 when `Features.batch_upload_v2_enabled?` is false
- [ ] When flag enabled, endpoints work as expected

### Direct Upload Authentication
- [ ] Custom `DirectUploadsController` inherits from `ActiveStorage::DirectUploadsController`
- [ ] Requires HMAC authentication (uses `Strata::Auth::Strategies::Hmac`)
- [ ] Returns 401 for missing/invalid HMAC signatures
- [ ] Returns presigned URL and headers for S3 upload when authenticated

### API Endpoints
- [ ] `POST /rails/active_storage/direct_uploads` returns presigned URL (HMAC auth)
- [ ] `POST /api/certification_batch_uploads` attaches blob and creates upload record (HMAC auth)
- [ ] `GET /api/certification_batch_uploads/:id` returns upload status (HMAC auth)
- [ ] All endpoints follow JSON API conventions

### Test Coverage
- [ ] DirectUploadsController requires valid HMAC signature
- [ ] API endpoints require HMAC auth
- [ ] Feature flag properly gates all endpoints
- [ ] Upload creation enqueues background job

## Implementation Notes

**DirectUploadsController**: Subclass `ActiveStorage::DirectUploadsController`, add `authenticate_api_request!` before_action using `Strata::Auth::Strategies::Hmac` (same pattern as `ApiController`).

**API Flow**:
1. Client calls `/rails/active_storage/direct_uploads` with HMAC auth → gets presigned URL
2. Client uploads file directly to S3
3. Client calls `/api/certification_batch_uploads` with blob signed_id → creates upload record and enqueues job

**Routes**: Override Active Storage's default direct_uploads route to use custom controller with HMAC auth.

## Points

**3 points** | Priority: **P1**

giverm:
Resolves #209

Adds three API endpoints for state system integrators to automate batch CSV submissions, all gated behind the batch_upload_v2 feature flag:

1. POST /api/direct_uploads — returns a presigned S3 upload URL (via ActiveStorage Direct Upload) with HMAC authentication
2. POST /api/certification_batch_uploads — creates a batch upload record from an uploaded blob and enqueues processing
3. GET /api/certification_batch_uploads/:id — poll upload status

Key design decisions:

- Api::DirectUploadsController inherits ActiveStorage::DirectUploadsController (ActionController::Base) rather than ApiController (ActionController::Metal), so HMAC auth is shared via an ApiHmacAuthentication concern rather than through class inheritance.

- The Devise monkey-patch on ActiveStorage::DirectUploadsController (from authenticated_active_storage.rb) is skipped via skip_before_action so HMAC auth runs instead of session auth.

- API uploads set uploader_id: nil and source_type: :api. A migration makes uploader_id nullable. CertificationBatchUploadPolicy scopes API clients to only see api-sourced records, while admins see all.

- Api::Client gains an admin? method (returns false) to prevent NoMethodError when policies delegate through the AdminPolicy chain.

- Staff views use safe navigation (uploader&.email || "API") to handle nil uploader on API-sourced uploads.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/certification_batch_upload.rb
+++ b/reporting-app/app/models/certification_batch_upload.rb
@@ -124,6 +124,20 @@ class CertificationBatchUpload < ApplicationRecord
     end
   end
 
+  # Returns true for v2 uploads, which store errors in normalized records
+  # rather than the results JSONB column.
+  #
+  # Heuristic: V2 uploads always have results: {} (set by check_completion!),
+  # while v1 uploads populate results with { successes: [...], errors: [...] }
+  # via complete_processing!.
+  #
+  # Note: Pending/processing uploads also have results: {} (the DB default),
+  # so this returns true for them as well. Callers in the view gate on
+  # completed? before branching on v2_upload?, so this is safe.
+  def v2_upload?
+    results.blank?
+  end
+
   # Check if can be processed
   def processable?
     pending?
--- a/reporting-app/app/controllers/staff/certification_batch_uploads_controller.rb
+++ b/reporting-app/app/controllers/staff/certification_batch_uploads_controller.rb
@@ -5,6 +5,7 @@ require "csv"
 module Staff
   class CertificationBatchUploadsController < AdminController
     self.authorization_resource = CertificationBatchUpload
+    MAX_DISPLAYED_ERRORS = 100
 
     before_action :set_batch_upload, only: [ :show, :process_batch, :results, :download_errors ]
 
@@ -29,6 +30,11 @@ module Staff
 
     # GET /staff/certification_batch_uploads/:id
     def show
+      @upload_errors = if @batch_upload.v2_upload?
+        @batch_upload.upload_errors.order(:row_number).limit(MAX_DISPLAYED_ERRORS)
+      else
+        []
+      end
     end
 
     # GET /staff/certification_batch_uploads/:id/results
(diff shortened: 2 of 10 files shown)


reporting-app/app/models/certification_batch_upload.rb
baonguyenNava: If we give Api::Client a UUID, I think we need to slightly alter this to return Api::Client if the uuid matches

PR #346, opened 2026-03-16
Create MemberOidcController and shared claim extracter

nkutub:
## Overview

Implements **Story 3** from the [Member SSO Implementation Stories](../architecture/staff-sso/member-sso-stories.md): a shared OIDC claim extractor and `Auth::MemberOidcController` so staff and member OIDC flows stay DRY and consistent.

**Story:** As a **developer**, I want a single place to extract OIDC claims from the OmniAuth auth hash and a member callback controller so that staff and member flows stay DRY and consistent.

### OidcClaimsExtractor

```ruby
# In controller
claims = extract_oidc_claims(auth, Rails.application.config.sso[:claims])
# or
claims = extract_oidc_claims(auth, Rails.application.config.member_oidc[:claims])
```

**Behavior:**
- Takes `(auth_hash, claim_config)` and returns `{ uid:, email:, name: }` plus optional `groups`, `region` when in config
- Staff uses `config.sso[:claims]`; member uses `config.member_oidc[:claims]` (uid, email, name only)
- Falls back to `auth.uid` when `unique_id` is missing from config
- Supports both string and symbol keys in raw_info

### Auth::MemberOidcController

Mirrors `Auth::SsoController` pattern:

| Action | Purpose |
|--------|---------|
| `new` | Renders auto-submit form that POSTs to OmniAuth `:member_oidc`; supports `origin` for deep links |
| `callback` | Extracts claims, calls `MemberOidcProvisioner.provision!`, signs in, redirects to `after_sign_in_path_for(user)` |
| `failure` | Redirects to `new_user_session_path` with generic alert |

**Security / robustness:**
- Skip `verify_authenticity_token` on callback only
- Skip `verify_authorized` and `verify_policy_scoped`
- Nil auth guard: if `request.env["omniauth.auth"]` is nil, log warning and redirect with failure message
- Rescue `ArgumentError` and `ActiveRecord::RecordInvalid` in callback; redirect with generic alert
- Failure action uses allowlisted `params[:message]` (invalid_credentials, timeout) to prevent log injection

### Auth::SsoController Refactor

- Replaced inline `extract_claims` with `extract_oidc_claims(auth, config.sso[:claims])`
- Added nil auth guard (matches member OIDC)
- Replaced raw `params[:message]` logging with `sanitized_failure_message` (allowlist)

## How to Test

```bash
bundle exec rspec spec/controllers/concerns/oidc_claims_extractor_spec.rb
bundle exec rspec spec/requests/auth/member_oidc_spec.rb
bundle exec rspec spec/requests/auth/sso_spec.rb  # Staff SSO still works
```

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/controllers/auth/member_oidc_controller.rb
+++ b/reporting-app/app/controllers/auth/member_oidc_controller.rb
@@ -61,19 +61,4 @@ class Auth::MemberOidcController < ApplicationController
   def member_oidc_enabled?
     Rails.application.config.member_oidc[:enabled]
   end
-
-  def redirect_if_authenticated
-    return unless user_signed_in?
-
-    redirect_to after_sign_in_path_for(current_user)
-  end
-
-  ALLOWED_OMNIAUTH_FAILURE_MESSAGES = %w[invalid_credentials timeout].freeze
-
-  def sanitized_failure_message(raw)
-    return "unknown_error" if raw.blank?
-
-    msg = raw.to_s.strip
-    ALLOWED_OMNIAUTH_FAILURE_MESSAGES.include?(msg) ? msg : "unknown_error"
-  end
 end
--- a/reporting-app/app/controllers/auth/sso_controller.rb
+++ b/reporting-app/app/controllers/auth/sso_controller.rb
@@ -85,19 +85,4 @@ class Auth::SsoController < ApplicationController
   def sso_enabled?
     Rails.application.config.sso[:enabled]
   end
-
-  def redirect_if_authenticated
-    return unless user_signed_in?
-
-    redirect_to after_sign_in_path_for(current_user)
-  end
-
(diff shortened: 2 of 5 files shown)


reporting-app/app/controllers/auth/member_oidc_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  ALLOWED_OMNIAUTH_FAILURE_MESSAGES = %w[invalid_credentials timeout].freeze
+
+  def sanitized_failure_message(raw)
+    return "unknown_error" if raw.blank?
+
+    msg = raw.to_s.strip
+    ALLOWED_OMNIAUTH_FAILURE_MESSAGES.include?(msg) ? msg : "unknown_error"
+  end
```
baonguyenNava: This looks identical to the methods in the sso_controller and looks like we can move these methods into the oidc_claims_extractor concern.

PR #420, opened 2026-04-01
IncomeService (create from intake)

jgavin-nava:
## Ticket

Resolves [OSCER-402](https://github.com/navapbc/oscer/issues/402)

## Changes

- Added `IncomeService` with `create_entry`, aligned with `ExParteActivityService`: returns a persisted `Income` on success, or `{ error: "..." }` on failure (no `:status` key; callers use `result[:error]` only, same pattern as `Certifications::CreationService` for hours). Duplicate detection is a **private** helper (unlike `ExParteActivityService`, which keeps `duplicate_entry?` public).
- `source_type` is passed straight to `Income`; invalid values are rejected by model validations (no separate normalization layer in the service).
- Optional `employer` is merged into `metadata` (string key `"employer"`); `reported_at` defaults to `Time.current` when omitted.
- Duplicate detection matches hours (member, category, gross income, period bounds)—`source_type` is not part of the duplicate key.
- Added `spec/services/income_service_spec.rb` covering success paths, duplicates, validation failures (including invalid `source_type` and category), and duplicate detection (via `send` for the private predicate).

## Context for reviewers

- Wiring `Certifications::MemberData::Activity` income fields and calling `IncomeService` from `Certifications::CreationService` is deferred until the VO and OpenAPI are extended for income-shaped activities (follow-up story).
- OpenAPI notes for optional `reported_at` at intake and API `source` → `source_type` belong with that schema work.

## Testing

- From `reporting-app/`: `make test` (full suite; passes SimpleCov thresholds).
- From `reporting-app/`: `make lint` (RuboCop clean).

Some console testing

```
IncomeService.create_entry(
  member_id: "test-member-1",
  category: "employment",
  gross_income: 580.00,
  period_start: Date.current.beginning_of_month,
  period_end: Date.current.end_of_month,
  source_type: "api"
)
# => #<Income ...> on success

IncomeService.create_entry(
  member_id: "test-member-1",
  category: "employment",
  gross_income: 580.00,
  period_start: Date.current.beginning_of_month,
  period_end: Date.current.end_of_month,
  source_type: "api"
)
# => {:error=>"Duplicate entry"}

IncomeService.create_entry(
  member_id: "test-member-2",
  category: "employment",
  gross_income: 580.00,
  period_start: Date.current.beginning_of_month,
  period_end: Date.current.end_of_month,
  source_type: "not_a_real_source"
)
# => {:error=>"Source type is not included in the list"}
```

♻️ Environment destroyed ♻️


reporting-app/app/services/income_service.rb
```diff
@@ -0,0 +1,80 @@
...
+    # @return [String, nil] canonical source_type value, or nil if unknown
+    def normalize_source_type(api_value)
+      return nil if api_value.nil?
+
+      raw = api_value.to_s.strip
+      return raw if Income::ALLOWED_SOURCE_TYPES.include?(raw)
+
+      key = raw.downcase.tr("-", "_").to_sym
+      Income::SOURCE_TYPES[key]
+    end
```
giverm: I wouldn't expect to need to normalize `source_type`. This wouldn't be a user submitted string, and the call sites should use the `Income::Source_TYPES` constant within the app.

giverm:
```suggestion

```

reporting-app/app/services/income_service.rb
```diff
@@ -0,0 +1,80 @@
...
+  class << self
+    # Create income data entry for a member
+    # @return [Income] on success
+    # @return [Hash] with :error, :status keys on failure
+    def create_entry(member_id:, category:, gross_income:, period_start:, period_end:,
+                     source_type:, source_id: nil, reported_at: Time.current, metadata: {}, employer: nil)
+      resolved_source_type = normalize_source_type(source_type)
+      unless resolved_source_type
+        return { error: "Source type is not included in the list", status: :unprocessable_entity }
+      end
```
giverm: Without normalization, we could just pass the source type directly and let the model validate


```suggestion
```

reporting-app/app/services/income_service.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      if duplicate_entry?(
+        member_id: member_id,
+        category: category,
+        gross_income: gross_income,
+        period_start: period_start,
+        period_end: period_end
+      )
+        return { error: "Duplicate entry", status: :conflict }
```
giverm: Looking at how `ExParteActivityService` is used, `:status` is unused at the call site (`Certifications::CreationService`). I think we should drop `:status` here, and follow up later to clean up `ExParteActivityService`.


```suggestion
        return { error: "Duplicate entry" }
```

reporting-app/app/services/income_service.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+
+      merged_metadata = merge_metadata(metadata, employer)
```
giverm: The `merged_metadata` variable is only used once, so I think we can just call the method in the new hash below rather that setting the variable here. 

Also, just curious, what sort of meta data are we expecting here? I don't see anything in the model about it.

reporting-app/app/services/income_service.rb
```diff
@@ -0,0 +1,80 @@
...
+    # @return [Boolean]
+    def duplicate_entry?(member_id:, category:, gross_income:, period_start:, period_end:)
+      Income.exists?(
+        member_id: member_id,
+        category: category,
+        gross_income: gross_income,
+        period_start: period_start,
+        period_end: period_end
+      )
+    end
```
giverm, line 52: Another nit for both services, should this be private?

reporting-app/app/services/income_service.rb
```diff
@@ -0,0 +1,80 @@
...
+        source_type: resolved_source_type,
+        source_id: source_id,
+        reported_at: reported_at,
+        metadata: merged_metadata
+      )
+
+      if entry.save
+        entry
+      else
+        { error: entry.errors.full_messages.join(", "), status: :unprocessable_entity }
```
giverm:
```suggestion
        { error: entry.errors.full_messages.join(", ") }
```

giverm (commented):
Looks good overall! Just a few comments/nits.

reporting-app/app/services/income_service.rb
jgavin-nava: That totally make sense

reporting-app/app/services/income_service.rb
jgavin-nava: employer is the only thing in the ticket that needs to be stored since there is not a dedicated column for it

reporting-app/app/services/income_service.rb
jgavin-nava: That is cool with me, I just kept it the same as the activity setup but I dont see the need for it being a public method, since it is only being called by the create entry

nkutub (approved):
LGTM

PR #427, opened 2026-04-09
Force reporting period step if no reporting period for activity report application form

### LINKED ISSUE #238: Bug: Member unable to select reporting period after exiting mid-flow
### Summary

When a member starts to report activities, they go through:

1. "Before you start" page
2. "Select reporting period" step

If the member exits at the "Select reporting period" step without selecting a period, and then returns to continue reporting later, they are not able to select a reporting period. The member should be taken back to the reporting period selection step since they have not completed it.

## User Story
As a member who started reporting activities but exited before selecting a reporting period, I need to be returned to the reporting period selection step when I resume, so that I can complete my activity reporting.

## Steps to Reproduce
You can also watch the video attached

1. Log in as a member
2. Start reporting activities
3. View "Before you start" page and proceed
4. Arrive at "Select reporting period" step
5. Exit or go to the dashboard without selecting a reporting period
6. Return to continue reporting
7. Observe: Member cannot select reporting period 

## Expected Behavior
Member should be taken back to the "Select reporting period" step to complete the selection before proceeding with activity reporting.

## Acceptance Criteria

- [ ] Member who exits before selecting reporting period is returned to that step on resume
- [ ] Member can successfully select a reporting period when returning
- [ ] Flow continues normally after reporting period selection

## Notes

- There is a longer term user research or heuristic evaluation on the back navigation needs that will be a separate ticket. For now, let's just send the member back to the reporting period step so they can select a reporting period. 

### Reproducibility

I tried multiple times and saw the bug every time

### How to reproduce the bug

https://github.com/user-attachments/assets/e2731d60-c259-4a9d-8b30-a0c62a16cb8b

### Browser

Chrome

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jdettmannnava:
## Ticket

Resolves [OSCER-238](https://github.com/navapbc/oscer/issues/238)
## Changes

- ActivityReportApplicationFormsController#show renders edit if no reporting periods

## Context for reviewers

Update all spec for /edit to use "Choose which month to report" instead of "Select reporting period" because "Select reporting period" in step header.

## Testing

`make lint` and `make test` pass

https://github.com/user-attachments/assets/423ec49a-2aab-46cb-9a0e-bc71872f6be6

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/controllers/activity_report_application_forms_controller.rb
+++ b/reporting-app/app/controllers/activity_report_application_forms_controller.rb
@@ -18,7 +18,7 @@ class ActivityReportApplicationFormsController < ApplicationController
 
   # GET /activity_report_application_forms/1 or /activity_report_application_forms/1.json
   def show
-    edit if @activity_report_application_form.reporting_periods.empty?
+    redirect_to edit_activity_report_application_form_path(@activity_report_application_form) if @activity_report_application_form.reporting_periods.empty?
   end
 
   # GET /activity_report_application_forms/new
@@ -57,7 +57,7 @@ class ActivityReportApplicationFormsController < ApplicationController
     @certification_requirements = @certification_case.certification.certification_requirements
 
     respond_to do |format|
-      format.html { render :edit }
+      format.html
       format.json { render json: @activity_report_application_form.as_json }
     end
   end
--- a/reporting-app/app/controllers/activity_report_application_forms_controller.rb
+++ b/reporting-app/app/controllers/activity_report_application_forms_controller.rb
@@ -18,6 +18,7 @@ class ActivityReportApplicationFormsController < ApplicationController
 
   # GET /activity_report_application_forms/1 or /activity_report_application_forms/1.json
   def show
+    edit if @activity_report_application_form.reporting_periods.empty?
   end
 
   # GET /activity_report_application_forms/new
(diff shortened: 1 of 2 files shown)


reporting-app/app/controllers/activity_report_application_forms_controller.rb
```diff
@@ -18,6 +18,7 @@ class ActivityReportApplicationFormsController < ApplicationController
 
   # GET /activity_report_application_forms/1 or /activity_report_application_forms/1.json
   def show
+    edit if @activity_report_application_form.reporting_periods.empty?
```
giverm: Should this be a redirect instead of calling the `edit` action directly? This will skip the before action, which currently works because show runs the same callback, but those could diverge. The redirect would also change the url from the show path to the edit path, which seems like what we'd want.

```suggestion
    if @activity_report_application_form.reporting_periods.empty?
      redirect_to edit_activity_report_application_form_path(@activity_report_application_form)
      return
    end
```

reporting-app/app/controllers/activity_report_application_forms_controller.rb
```diff
@@ -56,7 +57,7 @@ def edit
     @certification_requirements = @certification_case.certification.certification_requirements
 
     respond_to do |format|
-      format.html
+      format.html { render :edit }
```
giverm: This change can be reverted if you take the `redirect_to` suggestion

giverm (approved):
Looks good! Just one comment about redirecting vs. rendering.

giverm (approved):

PR #444, opened 2026-04-13
Business process — ex parte CE check (hours vs income)

jgavin-nava:
## Ticket

Resolves [[OSCER-406](https://github.com/navapbc/oscer/issues/406)](https://github.com/navapbc/oscer/issues/406)

## Changes

- **Ex parte CE system step** (`CertificationBusinessProcess::EX_PARTE_COMMUNITY_ENGAGEMENT_CHECK_STEP`) calls **`CommunityEngagementCheckService.determine`** (thin BP wiring only; no orchestration methods on the business process class).
- **`CommunityEngagementCheckService`:** Aggregates **hours** (`HoursComplianceDeterminationService.aggregate_hours_for_certification`) and **income** (`IncomeComplianceDeterminationService.aggregate_income_for_certification`), calls **`CertificationCase#record_ex_parte_ce_combined_assessment`**, then publishes **generic** community-engagement Strata events: **`DeterminedCommunityEngagementMet`**, **`DeterminedCommunityEngagementInsufficient`** (both fail with ex parte hours; payload includes `hours_data` and `income_data`), **`DeterminedCommunityEngagementActionRequired`** (both fail, no ex parte hours). BP transitions on this step use these names only (duplicate legacy `DeterminedHours*` transitions for this step removed).
- **Single combined determination** via `CertificationCase#record_ex_parte_ce_combined_assessment(certification:, …)`: callers pass the `Certification` aggregate. One automated `Determination` with `determination_data` type `Determination::CALCULATION_TYPE_EX_PARTE_CE_COMBINED`, nested `hours` / `income` payloads (each with `compliant`), and `satisfied_by` using **`Determination::SATISFIED_BY_*`** constants. **Outcome is compliant if either track meets its threshold; not compliant only when both fail.** When not compliant, **both** insufficient reason strings are stored (for compliant outcomes, reason codes reflect which track(s) passed).
- **`IncomeComplianceDeterminationService`:** Removed unused **`#determine`**. **`#calculate`** remains for silent income-only persistence (`record_income_compliance`, no Strata events). **`compliant_for_total_income?`** shared with `CommunityEngagementCheckService` and internal `determine_outcome`.
- **Specs:** `community_engagement_check_service_spec` covers combined recording and generic events (including **`DeterminedCommunityEngagementMet`** when both hours and income meet targets, and **combined `not_compliant` persistence** when both tracks fail with **no** ex parte hours / action-required). BP and model specs stub **`CommunityEngagementCheckService.determine`** during bootstrap (publishes **`DeterminedCommunityEngagementActionRequired`**) so tests do not accidentally persist a compliant CE determination on `CertificationCreated`; **`certification_case_spec`** also stubs **`NotificationService.send_email_notification`** so that bootstrap event does not run the mailer. `certification_case_spec` exercises `record_ex_parte_ce_combined_assessment` where applicable.
- **Docs:** `docs/architecture/income-data/income-data.md` — Business Process Integration step aligned with combined CE and generic events (step 4 split into sub-bullets for readability); **Testing / QE** paragraph (real `determine` vs stubbed bootstrap, OSCER-405 income stub, monitoring event names).
- **Follow-up:** `CommunityEngagementCheckService` — nil-safe `hours_data.dig(:hours_by_source, :ex_parte)` for the insufficient branch; `publish_workflow_events` takes `either_track_compliant:` instead of separate hour/income flags.

## Context for reviewers

- **Member status** still uses the latest determination’s `outcome`; OR semantics are encoded in that single row.
- **`NotificationsEventListener`** already subscribes to **`DeterminedCommunityEngagement*`** alongside legacy **`DeterminedHours*`** (other flows). Ex parte “insufficient” branch now drives **`insufficient_community_engagement_email`** (combined CE mailer), not **`insufficient_hours_email`**, for this step.
- **`HoursComplianceDeterminationService`** may still publish **`DeterminedHours*`** in other contexts; unrelated to the ex parte combined step’s event names.
- Use **`IncomeComplianceDeterminationService.calculate`** for silent income-only recalculation when applicable.
- **Idempotency:** The BP is expected to run the ex parte CE system step **once** per case for this transition; **`CommunityEngagementCheckService.determine` does not dedupe**—a second invocation would append another determination and publish events again. Confirm Strata/BP semantics if retries ever apply to the same case.
- **Ops / monitoring:** Include **`DeterminedCommunityEngagementMet`**, **`DeterminedCommunityEngagementInsufficient`**, and **`DeterminedCommunityEngagementActionRequired`** in log queries and dashboards for CE-related triage (in addition to legacy **`DeterminedHours*`** where those flows still run).

## Testing

- From `reporting-app/`: `make test` (full suite).
- Targeted: `make test args="spec/services/community_engagement_check_service_spec.rb spec/business_processes/certification_business_process_spec.rb spec/models/certification_case_spec.rb spec/services/income_compliance_determination_service_spec.rb spec/services/notifications_event_listener_spec.rb"`.
- `make lint` after edits.

### QE / product notes

- **Real `determine`:** `community_engagement_check_service_spec` (and listener specs that drive handlers) exercise **`CommunityEngagementCheckService.determine`** without the global bootstrap stub.
- **Stubbed CE on `CertificationCreated`:** `certification_case_spec` and `certification_business_process_spec` stub **`CommunityEngagementCheckService.determine`** (and related mailer stubs) so model/BP examples stay deterministic; they are **not** substitutes for the service spec’s integration coverage.
- **Income total for CE:** `member_reported_income_total` remains **`0` until OSCER-405**; combined CE threshold compares **`aggregate_income_for_certification`** totals (ex parte–sourced income path) only.

♻️ Environment destroyed ♻️


reporting-app/app/business_processes/certification_business_process.rb
```diff
@@ -18,8 +18,20 @@ class CertificationBusinessProcess < Strata::BusinessProcess
     ExemptionDeterminationService.determine(kase)
   })
 
+  # Ex parte CE check: hours path first, then income aggregate, else existing hours insufficient /
+  # action-required behavior (same Strata event names as today).
+  #
+  # Branching order (confirm with PM: hours-first vs parallel):
```
giverm: Has hours-first been confirmed?

reporting-app/app/business_processes/certification_business_process.rb
```diff
@@ -55,4 +69,28 @@ class CertificationBusinessProcess < Strata::BusinessProcess
...
+      HoursComplianceDeterminationService.determine(kase)
+    elsif income_compliant?(certification)
+      IncomeComplianceDeterminationService.determine(kase)
+    else
+      HoursComplianceDeterminationService.determine(kase)
+    end
+  end
+
+  def self.hours_compliant?(hours_data)
+    hours_data[:total_hours].to_f >= HoursComplianceDeterminationService::TARGET_HOURS
```
giverm: Could you add a `.compliant_for_total_hours?` method to `HoursComplianceDeterminationSerivce`  to keep the symmetry between the two compliance services?

giverm (approved):
Nice work!

reporting-app/app/business_processes/certification_business_process.rb
jgavin-nava: I added it to the hours cleanup ticket as I thought it might be out of scope for this specific one even though its really low risk it doesnt matter to me either way

reporting-app/app/business_processes/certification_business_process.rb
jgavin-nava: Nope you are right I meant to flip this its actually still in draft as its either income first or I am still researching if we are saving both and taking the complaint one if there is one. This is what happens when you dont re look at your notes from the day before

reporting-app/app/business_processes/certification_business_process.rb
giverm: Cool, that works for me, as long as it's documented

reporting-app/app/business_processes/certification_business_process.rb
```diff
@@ -55,4 +59,55 @@ class CertificationBusinessProcess < Strata::BusinessProcess
   transition(REPORT_ACTIVITIES_STEP, "ExemptionApplicationFormSubmitted", REVIEW_EXEMPTION_CLAIM_STEP)
   transition(REVIEW_EXEMPTION_CLAIM_STEP, "DeterminedExempt", END_STEP)
   transition(REVIEW_EXEMPTION_CLAIM_STEP, "DeterminedNotExempt", REPORT_ACTIVITIES_STEP)
+
+  # @param kase [CertificationCase]
+  def self.run_ex_parte_community_engagement_check(kase)
```
baonguyenNava: Not providing an in depth review, but broadly speaking, the business process is like an AWS step function. There shouldn't be these methods in here.

You can call a service, which updates the kase, followed by emitting an event to transition it to the next step.

baonguyenNava (dismissed):
Leaving a general comment about the structure of the business process.

Ideally, we may need to commit some rules about how a business process works, but I may have a hand in that soon.

reporting-app/app/business_processes/certification_business_process.rb
```diff
@@ -18,8 +18,12 @@ class CertificationBusinessProcess < Strata::BusinessProcess
     ExemptionDeterminationService.determine(kase)
   })
 
+  # Ex parte CE check: evaluates hours and income in parallel (both aggregated), stores one combined
+  # determination (+CertificationCase#record_ex_parte_ce_combined_assessment+). Member is compliant if
+  # either track meets its threshold; not compliant only if both fail. Same Strata event names as the
+  # legacy hours-only flow (+DeterminedHoursMet+ / +DeterminedHoursInsufficient+ / +DeterminedActionRequired+).
```
giverm: I maybe forgetting context I once had, but why are we still using the hours specific event names instead of using a more generalized event name?

reporting-app/app/models/certification_case.rb
```diff
@@ -205,4 +237,38 @@ def build_income_determination_data(income_data)
...
+
+  def build_ex_parte_ce_combined_determination_data(hours_data:, income_data:, hours_ok:, income_ok:)
+    satisfied_by = if hours_ok && income_ok
+      "both"
+    elsif hours_ok
+      "hours"
+    elsif income_ok
+      "income"
+    else
+      "neither"
```
giverm: I think these strings should be moved to `Determination` as class constants, since that's where they end up. I'm also not seeing any sort of defined structure for `Determination.determination_data`, which is a preexisting issue. I think going forward it would be a good idea to create a value object, similar to whats in `app/models/certifications/`. The value object is probably out of scope for this PR, but if you agee with that direction, maybe we could add an issue to the epic.

reporting-app/app/business_processes/certification_business_process.rb
```diff
@@ -55,4 +59,55 @@ class CertificationBusinessProcess < Strata::BusinessProcess
...
+
+    hours_ok = hours_compliant?(hours_data)
+    income_ok = IncomeComplianceDeterminationService.compliant_for_total_income?(income_data[:total_income])
+
+    kase.record_ex_parte_ce_combined_assessment(
+      hours_data: hours_data,
+      income_data: income_data,
+      hours_ok: hours_ok,
+      income_ok: income_ok
+    )
```
giverm: We should pass the certification here so we don't need to reload it inside the method
```suggestion
    kase.record_ex_parte_ce_combined_assessment(
      certification: certification,
      hours_data: hours_data,
      income_data: income_data,
      hours_ok: hours_ok,
      income_ok: income_ok
    )
```

reporting-app/app/models/certification_case.rb
```diff
@@ -145,6 +145,38 @@ def record_income_compliance(outcome, income_data)
...
+    outcome = (hours_ok || income_ok) ? :compliant : :not_compliant
+    reasons = ex_parte_ce_combined_reason_codes(outcome: outcome, hours_ok: hours_ok, income_ok: income_ok)
+    determination_data = build_ex_parte_ce_combined_determination_data(
+      hours_data: hours_data,
+      income_data: income_data,
+      hours_ok: hours_ok,
+      income_ok: income_ok
+    )
+
+    certification = Certification.find(certification_id)
```
giverm: We can remove this if certification is passed as an argument
```suggestion
```

reporting-app/app/services/income_compliance_determination_service.rb
giverm: Just curious, this method is currently unused, is there still a plan to wire up this method or have we changed direction?

reporting-app/app/business_processes/certification_business_process.rb
jgavin-nava: There would be still 2 different events once the income events are merged in. This was just using what is already in place. But I guess we could re work this a bit and just use one event for both of these. Once those are merged in though my thinking was 

If income or hours is complaint
  send the one that is or default to income
if both are not and have data
  go through both tracks
if one is and one is not 
  go through what has data

reporting-app/app/business_processes/certification_business_process.rb
jgavin-nava: Yea that makes sense no need to refind that case

reporting-app/app/models/certification_case.rb
jgavin-nava: Yea I think a defined structure would make sense at this point as we have an idea what we are looking at. I wonder if we made a decision to leave it unstructed becuase of the possibly differences in the activities but we might be at a good spot to define it

reporting-app/app/models/certification_case.rb
```diff
@@ -145,6 +146,38 @@ def record_income_compliance(outcome, income_data)
...
+      close! if outcome == :compliant
+
+      certification.record_determination!(
+        decision_method: :automated,
+        reasons: reasons,
+        outcome: outcome,
+        determination_data: determination_data,
+        determined_at: certification.certification_requirements.certification_date
+      )
+    end
```
giverm: Could this use `record_automated_ce_compliance` as well? I think you'd have to update the method to take a `:reasons` argument and pass the mapping rather than `compliant_reason` and `not_compliant_reason`.

reporting-app/spec/models/certification_case_spec.rb
```diff
@@ -5,6 +5,17 @@
 RSpec.describe CertificationCase, type: :model do
   let(:certification_case) { create(:certification_case) }
 
+  # Prevent real ex parte CE from recording a compliant determination during certification bootstrap
+  # (Income aggregate can meet threshold and close the case before examples run).
+  before do
```
giverm, line 10: We should add a stub so we don't actually run the mailer

```suggestion
  before do
    allow(NotificationService).to receive(:send_email_notifications)
```

giverm (approved):
LGTM! :shipit:

jgavin-nava:
Checked with @baonguyenNava this is fine to move forward since the feedback was addressed

PR #471, opened 2026-04-16
Fix Dependabot Rails group and ignore rule

giverm:
## Summary

Follow-up to #467. The `exclude-patterns` approach was cosmetic only — Dependabot still updated the Rails Gemfile constraint to 8.x inside the grouped PR (#470).

- **Replace `exclude-patterns` with a hard version ceiling** (`versions: [">=8.0.0"]`) that doesn't rely on Dependabot's semver classification, which is unreliable for Rails' 4-segment versioning
- **Add a `rails` group** (`rails`, `active*`, `action*`, `railties`) so all Rails sub-gems update together in a single PR, following the pattern used by Discourse and RubyGems.org

## Test plan

- [ ] After merge, close PR #470 (still contains Rails 8 bump) and verify Dependabot reopens a clean bundler PR without Rails 8
- [ ] Verify Rails sub-gems are grouped into their own PR on next Dependabot run

♻️ Environment destroyed ♻️

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,13 +23,19 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 3
+    ignore:
+      # Rails uses 4-segment versioning (e.g. 7.2.3.1) which causes Dependabot to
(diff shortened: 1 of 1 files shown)


jgavin-nava (approved):
LGTM

PR #494, opened 2026-04-21
 fix redirects app host docker

nkutub:
## Overview

Fixes incorrect **absolute redirect URLs** in environments where the app sees an **internal or container `Host`** header (e.g. Docker service name) instead of the browser-facing hostname. Rails builds many redirects as `request.protocol` + `request.host_with_port` + path, so users were sent to unreachable hosts after login or form flows.

When **`APP_HOST`** is set, the app now treats it as the **canonical public host** for Rack’s view of the request before controllers run, so `Location` headers match the URL users expect.

## Implementation notes

- **Why middleware:** `Rails.application.default_url_options` does not affect the Action Pack string redirect path (`redirect_to "/some/path"` → absolute URL from `request.host_with_port`). Normalizing Host / forwarded headers fixes all such redirects without auditing every controller.
- **Upstream headers:** If `HTTP_X_FORWARDED_PROTO` is already set (e.g. by a reverse proxy), it is not overwritten.
- **When `APP_HOST` is unset:** Middleware is a no-op; behavior matches previous releases.

## Configuration reminder

| Variable | Notes |
|----------|--------|
| `APP_HOST` | Public hostname; required in affected Docker / internal-routing setups for correct redirects when set. |
| `APP_PORT` | Optional; defaults and non-standard port handling mirror existing URL/OIDC conventions. |
| `DISABLE_HTTPS` | Use `true` only for HTTP public URLs; affects default `X-Forwarded-Proto` when unset. |

♻️ Environment destroyed ♻️


reporting-app/lib/middleware/public_request_host.rb
```diff
@@ -0,0 +1,50 @@
...
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      canonical = canonical_authority_from_env
+      if canonical
+        env["HTTP_HOST"] = canonical
+        env["HTTP_X_FORWARDED_HOST"] = canonical
+        env["HTTP_X_FORWARDED_PROTO"] ||= forwarded_proto_from_env
```
tomas-nava: you could add or modify a test to validate that these env values are set

reporting-app/lib/middleware/public_request_host.rb
```diff
@@ -0,0 +1,50 @@
...
+      @app.call(env)
+    end
+
+    private
+
+    def canonical_authority_from_env
+      app_host = ENV["APP_HOST"]&.strip
+      return nil if app_host.blank?
+
+      port = ENV.fetch("APP_PORT", "443").to_s.strip
```
tomas-nava, line 58: you could add or modify a test to validate that port is set to "443" when `ENV["APP_PORT"]` is blank

tomas-nava (commented):
This looks good to me, I left a couple of optional test suggestions.

I assume you're addressing this problem with middleware because it's more flexible than maintaining a list of trusted proxies in various container configurations/hosting environments?

Also, I think tests are failing in other areas of the app because the middleware is causing session resets; you could fix this by disabling the middleware when we're running in the test environment.

reporting-app/lib/middleware/public_request_host.rb
nkutub: added

reporting-app/lib/middleware/public_request_host.rb
nkutub: added

reporting-app/spec/lib/middleware/public_request_host_spec.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Middleware::PublicRequestHost do
+  def with_env(overrides, &block)
```
jgavin-nava: Might be outside the scope of this ticket but there is another spec `spec/requests/auth/sso_redirect_and_member_oidc_spec.rb` that redefines this same method. Maybe we should puil them out to a single helper

jgavin-nava (approved):

reporting-app/spec/lib/middleware/public_request_host_spec.rb
nkutub: I went ahead and extracted this to helper ..

nkutub:
> This looks good to me, I left a couple of optional test suggestions.
> 
> I assume you're addressing this problem with middleware because it's more flexible than maintaining a list of trusted proxies in various container configurations/hosting environments?
> 
> Also, I think tests are failing in other areas of the app because the middleware is causing session resets; you could fix this by disabling the middleware when we're running in the test environment.

actually this is related to Rails builds many redirects as request.protocol + request.host_with_port + path, so if Rack sees an internal Host, the Location header is wrong even when OIDC and default_url_options already assume a public APP_HOST.

As for the failing E2E test, this is due to Playwright was using the actual URL for the temporary environment setup but the middleware was redirecting off that original base url because it did not have that temp URL set in APP_HOST.

PR #497, opened 2026-04-22
Recalculation when income arrives for existing certifications

jgavin-nava:
## Ticket

Resolves [[OSCER-407](https://github.com/navapbc/oscer/issues/407)]

## Changes

- Added `CertificationCase.open_certification_id_for_member` to resolve the member’s latest open case’s `certification_id` for downstream compliance work.
- Extended `IncomeService#create_entry` with `recalculate_income_compliance` (default `true`): after a successful save, calls `IncomeComplianceDeterminationService.calculate` when an open case exists; rescues `ActiveRecord::RecordNotFound` and logs a warning.
- `Certifications::CreationService` passes `recalculate_income_compliance: false` during certification intake so income rows created before the case exists do not trigger the wrong or premature recalculation path.
- Documented the behavior in `docs/architecture/income-data/income-data.md` (compliance recalculation section and diagram edge).
- `record_automated_ce_compliance` supports **`close_on_compliant`**; `record_income_compliance` defaults **`true`** (same close-on-compliant behavior as hours; pass **`false`** later if product wants record-only recalculation).
- Specs: `income_service_spec`, `certification_case_spec`, `income_compliance_determination_service_spec`, `hours_compliance_determination_service_spec` for the new behavior, lookup, and silent-recalc parity.

## Context for reviewers

- **Issue wording vs implementation:** OSCER-407’s pseudocode mentioned possibly skipping `close!` when aligning with a future job; **product chose parity with `HoursComplianceDeterminationService#calculate`:** compliant silent recalculation **closes** the case via `CertificationCase#record_automated_ce_compliance` / `record_income_compliance(..., close_on_compliant: true)` (default). The keyword documents a future **record-only** mode without changing today’s behavior.
- Mirrors the silent `IncomeComplianceDeterminationService.calculate` contract (no Strata workflow events from that path). **`CalculateComplianceJob` is still not implemented** under `app/jobs/`; recalc runs **synchronously** from `IncomeService` after persist (hours service continues to reference the job only in comments).
- `member_id` on `Income` maps to certification via open cases only; newest certification wins if multiple open cases exist for the same member (edge case).

## Testing

- From `reporting-app/`: `make lint`
- From `reporting-app/`: `make test`
- Targeted: `make test args='spec/services/income_service_spec.rb spec/models/certification_case_spec.rb spec/services/income_compliance_determination_service_spec.rb spec/services/hours_compliance_determination_service_spec.rb'`

♻️ Environment destroyed ♻️


reporting-app/spec/services/income_service_spec.rb
```diff
@@ -53,6 +53,71 @@
       end
     end
 
+    context "when the member has an open certification case" do
+      let(:certification) { create(:certification) }
+      let!(:kase) { create(:certification_case, certification: certification) }
```
giverm: Prefer `let` over `let!` so we don't end up with unexpected objects in our test examples
```suggestion
      let(:kase) { create(:certification_case, certification: certification) }
```

reporting-app/app/models/certification_case.rb
```diff
@@ -27,6 +27,19 @@ class CertificationCase < Strata::Case
...
+  # @param member_id [String]
+  # @return [UUID, nil] certification_id
+  def self.open_certification_id_for_member(member_id)
+    joins("INNER JOIN certifications ON certifications.id = certification_cases.certification_id")
+      .where(certifications: { member_id: member_id })
+      .open
+      .order("certifications.created_at DESC")
+      .limit(1)
+      .pick(:certification_id)
+  end
```
giverm, line 48: Nit: I like using Arel over raw SQL strings for being composable and DB agnostic. Not critical for this case, up to you. 

```suggestion
  def self.open_certification_id_for_member(member_id)
    cases = arel_table
    certs = Certification.arel_table

    joins(
      cases.join(certs, Arel::Nodes::InnerJoin)
        .on(cases[:certification_id].eq(certs[:id]))
        .join_sources
    )
      .merge(Certification.where(member_id: member_id))
      .open
      .order(certs[:created_at].desc)
      .limit(1)
      .pick(:certification_id)
  end
```

giverm (approved):

reporting-app/app/models/certification_case.rb
jgavin-nava: I have never used arel before but that makes a lot of sense to use it for this project

reporting-app/spec/services/income_service_spec.rb
jgavin-nava: The reason I used let! here is because of the setup in the second test there is no explicit call to kase that I didn tthink it would create it but the test seem to be fine so I am good with removing the !

PR #515, opened 2026-04-28
OSCER-408: Staff case Activity Report income, compliance UI, and income aggregation fixes

jgavin-nava:
## Ticket

Resolves [[OSCER-408](https://github.com/navapbc/oscer/issues/408)](https://github.com/navapbc/oscer/issues/408)

## Changes

- **Staff case — Activity Report & External Data:** Compliance-style **hours** and **income** tables (aligned with OSCER-336 Figma direction), shared **`_hours_reported_table`** and **`_income_reported_table`** in both sections; **`_external_data_section`** (renamed from `_ex_parte_data_section`). Per-track **empty states** and **`hours_has_data` / `income_has_data`** gating so we do not show misleading “no hours” / “no income” copy when the other track has rows.
- **`CertificationCasesController#show`:** Loads external hourly / income in lookback; **`member_hour_activities_for_certification`** and **`member_income_activities_for_certification`** with **`certification_case: @case`**; **`aggregate_hours_for_certification`** and **`aggregate_income_for_certification`** with **prefetch kwargs** so staff show does not double-query. **`ApplicationHelper#state_name`** for `ENV.fetch("STATE_NAME", "the State")`.
- **`ActivityAggregator`:** **`certification_case_for_certification`** — deterministic tie-break when multiple cases share a **`certification_id`** (prefer case that owns an **`ActivityReportApplicationForm`**, else newest); debug log when ambiguous; **`fetch_member_activities`** uses it.
- **`HoursComplianceDeterminationService`:** **`aggregate_hours_for_certification`** accepts **`certification_case:`**, **`external_hourly_activities:`**, **`member_hour_activity_rows:`**; **`summarize_hours_for_aggregate`** for array inputs; **`member_hours_from_activities`** / **`reorder(nil)`** before SQL aggregation; **`calculate`** resolves case via shared helper. **`member_hour_activities_for_certification`** for staff “Hours reported” rows (`where.not(hours: nil)`).
- **`IncomeComplianceDeterminationService`:** Case resolution via **`certification_case_for_certification`** (removed duplicate private resolver); **`aggregate_income_for_certification`** prefetch kwargs; **`calculate`** uses shared case resolution.
- **`CommunityEngagementCheckService`**, **`CertificationCase`** (`accept_activity_report` / `deny_activity_report`), **`NotificationsEventListener`:** Pass **`certification_case:`** (or payload-derived case) into hours aggregation when recomputing so totals match the same case as income / staff UI. **`certification_case_for_notification`** validates **`certification_id`** on **`case_id`**.
- **`ExternalIncomeActivity`:** **`within_period`** lookback bounds; **locale contract spec** for **`certification_cases.income.source_types.*`**.
- **Locales:** **`certification_cases.external_data`**, **`certification_cases.income.*`**, **`certification_cases.common`**; no hard-coded English in the new compliance partials.
- **SCSS:** Compliance table styling; trimmed redundant **`!important`** where specificity suffices.
- **Docs:** `docs/reporting-app/staff-case-income-attribution-408.md` (implementation uses **`ExternalIncomeActivity`**, not the issue’s legacy **`Income.for_member`** sketch).
- **Specs:** `activity_aggregator_spec`, hours / income / notifications services, `certification_cases` request spec, external-income i18n contract.

### Scope note — please read first

[[OSCER-408](https://github.com/navapbc/oscer/issues/408)](https://github.com/navapbc/oscer/issues/408) **Story 8** in GitHub is written narrowly (controller fetch, income partial with locals, i18n). **This PR is intentionally broader** than that bullet list because:

1. **Product / design:** Staff hours and income are shown as **matching compliance tables** (same columns, footers, and empty-state rules). That required **shared partials** and **shared “has data” rules**, not a one-off income-only table.
2. **Correct data model:** Income in OSCER comes from **`ExternalIncomeActivity`** (and member report line items), not **`Income.for_member`** as in the old issue text — the PR matches production.
3. **Trustworthy numbers next to the UI:** Once income and hours use the same layout, **aggregation must use the same `CertificationCase`** as the page (multi-case edge) and **avoid redundant queries** (prefetch pattern). That pulled in **`ActivityAggregator#certification_case_for_certification`**, hours **`aggregate_*` kwargs**, CE / notification / accept-deny call sites, and **income service deduplication** of case resolution.

**If you prefer to mentally “split” the diff:** treat **(A)** staff views + i18n + shared partials as the **visible 408** outcome, and **(B)** aggregator / hours+income services / CE+notifications+`CertificationCase` as **correctness and parity** required so the UI does not lie or drift. It is a lot of files, but most paths are **tightly coupled** to “same tables, same case, same totals.”

**Review focus suggestions**

- **Staff UX:** `certification_cases/show`, `_external_data_section`, `_hours_reported_table`, `_income_reported_table`, locale keys under **`certification_cases.*`**.
- **Correctness / edge cases:** `certification_case_for_certification` tie-break; **`aggregate_*`** with and without prefetch kwargs; notification payload **`case_id`** vs certification mismatch guard.
- **Risk:** Any caller of **`aggregate_hours_for_certification`** / **`aggregate_income_for_certification`** that should pin a case but omits **`certification_case:`** — call that out if you spot one.

### Other context

- From **`reporting-app/`:** `make lint` or `make lint-ci`, then **`make test`** (full suite before merge).
- Targeted examples: `spec/services/concerns/activity_aggregator_spec.rb`, `spec/services/hours_compliance_determination_service_spec.rb`, `spec/services/income_compliance_determination_service_spec.rb`, `spec/services/notifications_event_listener_spec.rb`, `spec/requests/certification_cases_spec.rb`.

## Testing

- **Before merge:** From `reporting-app/`, run **`make lint`** (or **`make lint-ci`**) and **`make test`** (full suite).
- **Prior runs:** `make lint-ci` plus targeted specs for hours + certification cases requests passed; full suite was green on an earlier branch tip. SimpleCov may exit **2** on single-file / narrow runs — use full **`make test`** for CI parity.

Screenshots

Screenshot 2026-05-14 at 11 49 41 AM
Screenshot 2026-04-29 at 1 34 12 PM
Screenshot 2026-05-14 at 11 40 39 AM

♻️ Environment destroyed ♻️


reporting-app/app/assets/stylesheets/application.scss
```diff
@@ -20,6 +20,56 @@ legend {
...
+    border-left: none !important;
+    border-right: none !important;
+    border-bottom: none !important;
+  }
+
+  tfoot tr:nth-child(2) td {
+    border: none !important;
+  }
+
+  tfoot tr:nth-child(3) td {
```
jdettmannnava, line 120: It looks like there are not always three trs in the footer, so should this be last-child?

reporting-app/app/views/certification_cases/_hours_reported_table.html.erb
```diff
@@ -1,58 +1,79 @@
 <% total_hours = hours_summary[:total_hours] %>
 <% hours_needed = [target_hours - total_hours, 0].max %>
 <% has_activities = ex_parte_activities.any? || member_activities.any? %>
+<% compliance = local_assigns.fetch(:compliance_table_styles, false) %>
```
jdettmannnava: Can't most of these style differences be made in the css file?
 I think they should be there instead of in the code if possible.

reporting-app/app/views/certification_cases/_hours_reported_table.html.erb
```diff
@@ -1,58 +1,79 @@
...
-  <table class="usa-table usa-table--borderless width-full">
+  <h3 class="<%= "margin-top-0 font-sans-lg" if compliance %>"><%= t(".title") %></h3>
+  <table class="<%= table_class %>">
     <thead>
       <tr>
-        <th scope="col"><%= t(".organization_name") %></th>
-        <th scope="col"><%= t(".source") %></th>
-        <th scope="col"><%= t(".activity_type") %></th>
-        <th scope="col" class="text-right"><%= t(".hours_reported") %></th>
+        <th scope="col"<% if compliance %> class="bg-white"<% end %>><%= t(".organization_name") %></th>
```
jdettmannnava: I think this can be in the css file under table.certification-case-compliance-table { thead th {...

reporting-app/app/views/certification_cases/_hours_reported_table.html.erb
```diff
@@ -1,58 +1,79 @@
 <% total_hours = hours_summary[:total_hours] %>
 <% hours_needed = [target_hours - total_hours, 0].max %>
 <% has_activities = ex_parte_activities.any? || member_activities.any? %>
+<% compliance = local_assigns.fetch(:compliance_table_styles, false) %>
+<% body_cell = compliance ? "border-0" : nil %>
+<% value_cell = compliance ? "border-0 text-right bg-base-lightest" : "text-right" %>
```
jdettmannnava: You might need to add a custom class for this in the scss, then style it when under +table.certification-case-compliance-table.
Also true for th_value below

reporting-app/app/views/certification_cases/_hours_reported_table.html.erb
```diff
@@ -1,58 +1,79 @@
...
-      </tr>
-      <tr>
-        <td colspan="3"><%= t(".required") %></td>
-        <td class="text-right"><%= target_hours %></td>
-      </tr>
-      <% if hours_needed > 0 %>
-        <tr class="bg-base-lightest">
-          <td colspan="3" class="text-bold"><%= t(".additional_hours_needed") %></td>
-          <td class="text-right text-bold"><%= number_with_precision(hours_needed, precision: 0) %></td>
+      <% if compliance %>
```
jdettmannnava: I think this flow control can be replaced by changes in the css file.

reporting-app/app/services/income_compliance_determination_service.rb
```diff
@@ -38,13 +39,15 @@ def calculate(certification_id)
     # Same lookback and query shape as ActivityAggregator#fetch_ex_parte_activities /
     # Income.for_member(...).within_period(lookback) as used for ex parte hours parity.
     # @param certification [Certification]
+    # @param certification_case [CertificationCase, nil] When nil, resolves the case that owns the activity report (if any) so member income is not read from the wrong row when multiple cases share a certification_id (e.g. test factories).
     # @return [Hash]
-    def aggregate_income_for_certification(certification)
+    def aggregate_income_for_certification(certification, certification_case: nil)
```
giverm: Now that we have `certification_case` as an argument, we should check the call sites to see where we can pass the case and avoid double look up, or potentially return results for a different case. `CommunitityEngagementCheckService.determine(kase)` is one instance I noticed that could now pass the case to `aggregate_income_for_certification`

reporting-app/app/views/certification_cases/_hours_reported_table.html.erb
```diff
@@ -1,58 +1,79 @@
...
+          <tr class="bg-base-lightest">
+            <td colspan="3" class="text-bold"><%= t(".additional_hours_needed") %></td>
+            <td class="text-right text-bold"><%= number_with_precision(hours_needed, precision: 0) %></td>
+          </tr>
+        <% end %>
       <% end %>
     </tfoot>
   </table>
-<% else %>
+<% elsif !local_assigns.fetch(:income_has_data, false) %>
```
giverm: This now couples the hours table to income. I think the empty table check should actually be the responsibility of the parent view rather than within this table partial.

reporting-app/app/views/certification_cases/_hours_reported_table.html.erb
```diff
@@ -1,58 +1,79 @@
 <% total_hours = hours_summary[:total_hours] %>
 <% hours_needed = [target_hours - total_hours, 0].max %>
 <% has_activities = ex_parte_activities.any? || member_activities.any? %>
+<% compliance = local_assigns.fetch(:compliance_table_styles, false) %>
+<% body_cell = compliance ? "border-0" : nil %>
+<% value_cell = compliance ? "border-0 text-right bg-base-lightest" : "text-right" %>
+<% th_value = compliance ? "text-right bg-base-lightest width-card" : "text-right" %>
+<% table_class = compliance ? "usa-table width-full certification-case-compliance-table" : "usa-table usa-table--borderless width-full" %>
```
giverm: All these conditionals, and the branching logic in `<tfoot>` make me wonder if these should be two separate partials. We'd end up with some duplication, but I think it'd be easier to read and maintain. We could also possibly simplify this partial by creating a partial for just the row, but that may just end up pushing the complexity down a layer.

reporting-app/config/locales/views/certification_cases/en.yml
```diff
@@ -35,6 +39,23 @@ en:
       step: "Step:"
       tasks: "Tasks"
       title: "Case: %{case_number}"
+    income:
+      table_title: "Income reported"
+      income_reported_amount: "Income reported"
+      total_reported: "Total reported"
+      required: "Required"
+      additional_income_needed: "Additional income needed"
+      no_rows_compliance: "No income reported for this lookback."
```
jdettmannnava:
```suggestion
      no_reported_income: "No income reported for this lookback."
```

reporting-app/config/locales/views/certification_cases/en.yml
```diff
@@ -35,6 +39,23 @@ en:
...
+      organization_name: "Organization name"
+      organization_not_listed: "—"
+      pay_period: "Pay period"
+      source: "Source"
+      activity_type: "Activity type"
+      gross_income: "Gross income"
+      no_rows: "No income data from external sources for this lookback."
```
jdettmannnava:
```suggestion
      no_ex_parte_income: "No income data from external sources for this lookback."
```

reporting-app/app/views/certification_cases/_incomes_table.html.erb
```diff
@@ -0,0 +1,27 @@
+<h3 class="margin-top-4"><%= title %></h3>
```
jdettmannnava: I think we should either have separate tables for reported and ex parte hours, or this should be combined with _income_reported_table.
If we keep them separate, we should rename this table to _income_ex_parte_table

reporting-app/app/views/certification_cases/_incomes_table.html.erb
```diff
@@ -0,0 +1,27 @@
...
+      <% income_rows.each do |row| %>
+        <tr>
+          <td><%= row.metadata&.dig("employer").presence || t("certification_cases.income.organization_not_listed") %></td>
+          <td class="text-no-wrap"><%= l(row.period_start) %> – <%= l(row.period_end) %></td>
+          <td><%= t("certification_cases.income.source_types.#{row.source_type}", default: row.source_type.to_s.tr("_", " ").titleize) %></td>
+          <td><%= row.category.titleize %></td>
+          <td class="text-right"><%= number_to_currency(row.gross_income) %></td>
+        </tr>
+      <% end %>
+    </tbody>
```
jdettmannnava: I don't have access to the figma, but it feels weird not to have a rollup here, although it has one for hours.

reporting-app/app/services/income_compliance_determination_service.rb
jgavin-nava: Yea that makes sense. I made sure that there wasnt double lookups for this implementation but should go back and check to find other examples that we can fix too

reporting-app/app/controllers/certification_cases_controller.rb
```diff
@@ -28,6 +28,16 @@ def show
     @target_hours = HoursComplianceDeterminationService::TARGET_HOURS
     @ex_parte_activities = fetch_ex_parte_activities
     @member_activities = fetch_member_activities
```
jdettmannnava: Since these are just hours activities, I think this variable (and method) should indicate that.

reporting-app/app/views/certification_cases/_hours_reported_table.html.erb
jgavin-nava: Yea that totally makes sense

jdettmannnava (commented):
Looks great!
But I think some of the view logic around css classes can be simplified.

reporting-app/app/assets/stylesheets/application.scss
jgavin-nava: No because the dark top border is only for the optional “Additional hours / income needed” row. Which would always be the 3rd child if we made it last child then this effect would bump up when the 3rd row is not there

reporting-app/app/controllers/certification_cases_controller.rb
```diff
@@ -24,10 +24,32 @@ def show
     @tasks = @case.tasks.order(created_at: :desc)
 
     # Hours data for the "Hours reported" table
-    @hours_summary = HoursComplianceDeterminationService.aggregate_hours_for_certification(@certification)
     @target_hours = HoursComplianceDeterminationService::TARGET_HOURS
     @external_hourly_activities = fetch_external_hourly_activities
-    @member_activities = fetch_member_activities
+    @hours_member_activities = HoursComplianceDeterminationService.member_hour_activities_for_certification(
```
jdettmannnava:
```suggestion
    @member_hours_activities = HoursComplianceDeterminationService.member_hour_activities_for_certification(
```
matches income below and makes more sense to me

reporting-app/app/controllers/certification_cases_controller.rb
```diff
@@ -24,10 +24,32 @@ def show
...
+      @certification,
+      certification_case: @case
+    )
+    hours_member_rows = @hours_member_activities.to_a
```
jdettmannnava:
```suggestion
    member_hours_rows = @member_hours_activities.to_a
```

reporting-app/app/services/hours_compliance_determination_service.rb
```diff
@@ -53,24 +53,42 @@ def determine_after_activity_report(kase)
...
-
-      external_hours = summarize_hours(external_hourly_activities)
-      member_hours = summarize_hours(member_activities)
+    def aggregate_hours_for_certification(
+      certification,
+      certification_case: nil,
+      external_hourly_activities: nil,
+      member_hour_activity_rows: nil
+    )
+      external_source = external_hourly_activities.nil? ? fetch_external_hourly_activities(certification) : external_hourly_activities
```
jdettmannnava:
```suggestion
      external_sources = external_hourly_activities.nil? ? fetch_external_hourly_activities(certification) : external_hourly_activities
```
It's plural...

reporting-app/app/services/income_compliance_determination_service.rb
```diff
@@ -42,13 +43,31 @@ def calculate(certification_id)
...
-    def aggregate_income_for_certification(certification)
+    def aggregate_income_for_certification(
+      certification,
+      certification_case: nil,
+      external_income_activities: nil,
+      member_income_activity_rows: nil
+    )
       lookback_period = certification.certification_requirements.continuous_lookback_period
-      external_income_activities = fetch_external_income_activities(certification, lookback_period)
+      external_source = if external_income_activities.nil?
```
jdettmannnava:
```suggestion
      external_sources = if external_income_activities.nil?
```

reporting-app/app/controllers/certification_cases_controller.rb
giverm: Can we align the variable names? We have `@hours_member_activities` here, and below it's `@member_income_activities`. I have a slight preference for having hours/income first.

reporting-app/app/services/income_compliance_determination_service.rb
```diff
@@ -63,21 +82,64 @@ def aggregate_income_for_certification(certification)
...
       compliant_for_total_income?(total_income) ? :compliant : :not_compliant
     end
 
-    # Approved member-reported income (e.g. from activity report) — stub until modeled.
-    # @param _certification [Certification]
-    # @return [BigDecimal]
-    def member_reported_income(_certification)
-      # TODO(OSCER-405): Sum approved member income activities when that model/workflow exists.
+    def summarize_income_for_aggregate(activities)
+      return summarize_income(activities) if activities.is_a?(ActiveRecord::Relation)
```
jdettmannnava: Same as `summarize_hours` comment above

reporting-app/app/views/certification_cases/_external_data_section.html.erb
```diff
@@ -0,0 +1,31 @@
+<% if activity_report.present? %>
+  <p class="maxw-full"><%= t("certification_cases.external_data.summarized_under_activity_report") %></p>
+<% else %>
+  <% hours_has_data = external_hourly_activities.any? || hours_member_activities.any? %>
+  <% income_member_activities = local_assigns.fetch(:income_member_activities) %>
+  <% income_has_data = external_income_activities.any? || income_member_activities.any? %>
```
giverm: `hours_has_data` and `income_has_data` are also calculated in `show.html` on lines 110 and 111. Could we pass those values here, or calculate them in the controller?

reporting-app/app/views/certification_cases/_income_reported_table.html.erb
```diff
@@ -0,0 +1,55 @@
...
+    <tbody>
+      <% external_income_activities.each do |row| %>
+        <% category_name = row.category.titleize %>
+        <tr>
+          <td>
+            <%= row.metadata&.dig("employer").presence || "#{category_name} #{t('certification_cases.common.activity_label')}" %>
+          </td>
+          <td><%= t("certification_cases.common.source_external", state_name: state_name) %></td>
+          <td><%= category_name %></td>
+          <td><%= number_to_currency(row.gross_income) %></td>
```
giverm: Should this be `precision: 0` to match other currency amount being displayed on the table?

reporting-app/app/services/concerns/activity_aggregator.rb
```diff
@@ -24,6 +24,30 @@ def fetch_member_activities(certification)
...
+  # +created_at+); otherwise falls back to the newest case. Single source of truth for the
+  # tie-break shared by +HoursComplianceDeterminationService+ and +IncomeComplianceDeterminationService+.
+  # @param certification [Certification]
+  # @param certification_case [CertificationCase, nil]
+  # @return [CertificationCase, nil]
+  def certification_case_for_certification(certification, certification_case = nil)
+    return certification_case if certification_case
+
+    scoped = CertificationCase.where(certification_id: certification.id)
+    if scoped.offset(1).exists?
```
jdettmannnava: I think it can just return first if length < 2.
```suggestion
    return scoped.first unless scoped.length > 1

    if scoped.offset(1).exists? # and also remove this if
```
Also, I don't like the name `scoped` here, as `scope` (and `with` below) have other meanings, but I also don't like arguing about variable names...

reporting-app/app/services/hours_compliance_determination_service.rb
```diff
@@ -126,5 +162,24 @@ def merge_category_hours(external, member)
         result[category] = (external[category] || 0.0) + (member[category] || 0.0)
       end
     end
+
+    def summarize_hours_for_aggregate(activities)
```
jdettmannnava: I think this logic can move to `summarize_hours` (I believe it is only used here?), and we can get rid of this method.

giverm (approved):
Looks good!

reporting-app/app/services/notifications_event_listener.rb
```diff
@@ -128,6 +142,18 @@ def fetch_certification(event)
...
     end
 
+    # When recomputing aggregates for a notification, prefer the case from the event payload so member
+    # income matches the case under workflow (same as CommunityEngagementCheckService / staff show).
+    def certification_case_for_notification(certification, payload)
+      case_id = payload[:case_id]
+      return nil if case_id.blank?
+
+      kase = CertificationCase.find_by(id: case_id)
+      return nil if kase.blank? || kase.certification_id != certification.id
```
jdettmannnava:
```suggestion
      return nil if kase&.certification_id != certification.id
```

reporting-app/app/views/certification_cases/_hours_reported_table.html.erb
```diff
@@ -1,58 +1,58 @@
...
+  <tbody>
+    <% external_hourly_activities.each do |activity| %>
+      <% category_name = activity.category.titleize %>
       <tr>
-        <th scope="col"><%= t(".organization_name") %></th>
-        <th scope="col"><%= t(".source") %></th>
-        <th scope="col"><%= t(".activity_type") %></th>
-        <th scope="col" class="text-right"><%= t(".hours_reported") %></th>
+        <td><%= category_name %> <%= t("#{common}.activity_label") %></td>
+        <td><%= t("#{common}.source_external", state_name: state_name) %></td>
```
jdettmannnava:
```suggestion
        <td><%= t("#{common}.source_external", state_name:) %></td>
```

reporting-app/app/views/certification_cases/_income_reported_table.html.erb
```diff
@@ -0,0 +1,55 @@
...
+      </tr>
+    </thead>
+    <tbody>
+      <% external_income_activities.each do |row| %>
+        <% category_name = row.category.titleize %>
+        <tr>
+          <td>
+            <%= row.metadata&.dig("employer").presence || "#{category_name} #{t('certification_cases.common.activity_label')}" %>
+          </td>
+          <td><%= t("certification_cases.common.source_external", state_name: state_name) %></td>
```
jdettmannnava:
```suggestion
          <td><%= t("certification_cases.common.source_external", state_name:) %></td>
```

jdettmannnava (commented):
Looks great! Some refactoring suggestions

reporting-app/app/controllers/certification_cases_controller.rb
jgavin-nava: Yea that definitely should be the same

jgavin-nava: Hahaha looks like you and mike are right on the same page definitely just an oversight will make that change

reporting-app/app/views/certification_cases/_hours_reported_table.html.erb
jgavin-nava: TIL

reporting-app/app/views/certification_cases/_income_reported_table.html.erb
jgavin-nava: I am thinking now that the currency should not have this precisison as it removes the cents but not sure

reporting-app/app/services/concerns/activity_aggregator.rb
jgavin-nava: The variable names make sense to change make them at least more meaningful.

For the other part. This is just logging at the debug level incase there are 2 with the same id. But then continues with the code and returns the first one so I think it is unnecessary to return first early on. I can remove the log message if that seems like overklill though

jdettmannnava: Isn't it also making another database call to check for ActivityReportApplicationForm?

jgavin-nava: Ohh I see what you mean changing the logic because the extra queries are unnecessary if there is already only 1 certification case which is most likely. And only do the extra queries when necessary. That absolutely makes sense now

reporting-app/app/controllers/certification_cases_controller.rb
jgavin-nava: Changed the income section from the suggestion from mike to keep hours/income first here

reporting-app/app/services/concerns/activity_aggregator.rb
```diff
@@ -24,6 +24,31 @@ def fetch_member_activities(certification)
...
+  # +created_at+); otherwise falls back to the newest case. Single source of truth for the
+  # tie-break shared by +HoursComplianceDeterminationService+ and +IncomeComplianceDeterminationService+.
+  # @param certification [Certification]
+  # @param certification_case [CertificationCase, nil]
+  # @return [CertificationCase, nil]
+  def certification_case_for_certification(certification, certification_case = nil)
+    return certification_case if certification_case
+
+    cc_scoped = CertificationCase.where(certification_id: certification.id)
+    cases = cc_scoped.limit(2).to_a
```
jdettmannnava: I don't think this is necessary?

jgavin-nava: Yea it is probably overkill again. I was trying to limit the size of a call to the database if there are bunch of cases that have the same cert id so this would make the performance be way down if that is the case but I think it makes sense that it is not going to happen very often and if anything would be in testing

reporting-app/app/controllers/certification_cases_controller.rb
```diff
@@ -24,10 +24,34 @@ def show
...
+    @hours_summary = HoursComplianceDeterminationService.aggregate_hours_for_certification(
+      @certification,
+      certification_case: @case,
+      external_hourly_activities: @external_hourly_activities,
+      member_hour_activity_rows: hours_member_rows
+    )
+    @external_income_activities = fetch_external_income_activities
+    @member_income_activities = IncomeComplianceDeterminationService.member_income_activities_for_certification(
```
jdettmannnava, line 41: Shouldn't this be income_member_activities to match hours_member_activities above?

jgavin-nava: Yea thought I got them all thank you though. Now that I look at this the method attached to this should probably change the name as well

jdettmannnava (approved):
LGTM!

PR #583, opened 2026-05-21
OSCER-571 log decisions

### LINKED ISSUE #571: CE Write case management events to Audit Log
## Background
To meet compliance with CMS, OSCER must log all CE decisions to a read-only log.

## Acceptance Criteria
All CE decisions should be written to the Strata::AuditLog
- Accept/Deny CE based on external exemption information
- Accept/Deny CE based on external activity information
- Accept/Deny CE based on exemption information provided by member
- Accept/Deny CE based on activity information provided by member
- Task assignment

Each log must provide
- The action
- Who did it (specific user or process)
- Timestamp

Should use [Strata::AuditLog](https://github.com/navapbc/strata-sdk-rails/blob/main/docs/strata-audit-log.md)

jdettmannnava:
## Ticket

Resolves #571 

## Changes

- Determinable#record_determination! logs event to AuditLog
- Decisions without determinations (denied exemptions) log directly to AuditLog
- Task assignments logged to AuditLog
- Added `actor` parameter to CertificationCase methods that record decisions

## Context for reviewers

We want all decisions regarding Community Engagement Compliance to be logged in the AuditLog, which (unlike Determinations) are immutable.

The `actor` parameter is anything the AuditLog can use, so an ActiveRecord object or a Strata::VirtualActor. Because of this requirement, determinations get determined_by_id for free.

## Testing

Automated tests passed.
Manually created the following and verified the expected results in the AuditLog (via Rails console):
- Automated pass exemption
- Automated pass hours
- Automated pass income
- Assign task by pick next
- Assign task directly
- Deny Exemption
- Deny Activity Report
- Accept Exemption
- Accept Activity Report

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/controllers/tasks_controller.rb
+++ b/reporting-app/app/controllers/tasks_controller.rb
@@ -28,8 +28,16 @@ class TasksController < Strata::TasksController
 
   def assign
     set_task
-    @task.assign(current_user.id)
-    flash["task-message"] = "Task assigned to you."
+    Strata::AuditLog.record(actor: current_user) do |log|
+      @task.assign(current_user.id)
+
+      log.add_line(
+        action: "case.task_picked_up",
+        subject: Certification.find(@task.case.certification_id),
+        data: { task_id: @task.id, task_type: @task.type }
+      )
+    end
+    flash["task-message"] = I18n.t("strata.tasks.messages.task_picked_up")
     redirect_to task_path(@task)
   end
 
@@ -41,12 +49,19 @@ class TasksController < Strata::TasksController
     # Default scope on Strata::Task is ordered by :due_on
     task = policy_scope(Strata::Task).incomplete.unassigned.first
 
-    if task && task.assign(current_user.id)
-      flash["task-message"] = I18n.t("strata.tasks.messages.task_picked_up")
-      redirect_to task_path(task)
-    else
-      flash["task-message"] = I18n.t("strata.tasks.messages.no_tasks_available")
-      redirect_to tasks_path
+    Strata::AuditLog.record(actor: current_user) do |log|
+      if task && task.assign(current_user.id)
+        log.add_line(
+          action: "case.task_picked_up",
+          subject: Certification.find(task.case.certification_id),
+          data: { task_id: task.id, task_type: task.type }
+        )
+        flash["task-message"] = I18n.t("strata.tasks.messages.task_picked_up")
+        redirect_to task_path(task)
+      else
+        flash["task-message"] = I18n.t("strata.tasks.messages.no_tasks_available")
+        redirect_to tasks_path
+      end
     end
   end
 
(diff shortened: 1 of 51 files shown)


reporting-app/app/controllers/tasks_controller.rb
```diff
@@ -28,8 +28,16 @@ def index
 
   def assign
     set_task
-    @task.assign(current_user.id)
-    flash["task-message"] = "Task assigned to you."
+    Strata::AuditLog.record(actor: current_user) do |log|
+      @task.assign(current_user.id)
```
jgavin-nava: I think it would make sense to make sure assign succeeds before adding the log

reporting-app/app/services/exemption_determination_service.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class ExemptionDeterminationService
+  include Strata::VirtualActor
```
jgavin-nava, line 4: Do we want to add this same thing to the hours and income compliance_dertmination_service? Feels like they would be virtual actors and log as self

reporting-app/app/models/concerns/determinable.rb
```diff
@@ -47,4 +47,25 @@ module Determinable
...
+      reasons:,
+      outcome:,
+      determination_data:,
+      determined_at:,
+      determined_by_id:
+    )
+
+    determination_method = (reasons & Determination::EXEMPTION_REASONS).any? ? :exemption : :activity_report
+    determination_status = outcome.to_sym == :not_compliant ? :denied : :approved
+    Strata::AuditLog.write!(
```
jgavin-nava, line 63: I was reading about maybe usinjg the .record method here and bringing the super command into so if it falls it would be wrapped inside of it and would populate on success if I am understanding it correctly.

jdettmannnava: I started doing it that way but decided against it mostly because we need to return the determination to meet the contract. Scoping a nil determination outside the scope didn't seem worth it, as we always call record_determination! in a transaction already, and record_determination! itself throws an error on fail so we shouldn't get any false audit logs.

reporting-app/app/services/exemption_determination_service.rb
jdettmannnava: income_compliance_determination_service never records determination, so it doesn't need this.
As for hours_compliance_determination_service, I don't think we call `determine` on it since the consolidated determination work. Is this your understanding as well? I didn't want to bother adding this code there, as it is unused, and I didn't want to clog the PR with removing the `determine` method.
I can make a ticket to remove this method if you agree that it is no longer used, and/or add this functionality to the method if you think it should be there.

jgavin-nava: Ohh that totally makes sense. I think we can make that ticket

jdettmannnava: https://github.com/navapbc/oscer/issues/588

reporting-app/app/models/concerns/determinable.rb
jgavin-nava: Yea that definitely makes sense now. Just wanted to bring it up and see your thoughts. Glad it was already taken into account.

jgavin-nava (approved):

Issue #585, opened 2026-05-22
Verification Window: Bind review tasks directly to their application form


jdettmannnava:
## Background

Members will need to be able to re-submit forms after staff mark the case as not compliant. This behavior is currently blocked by uniqueness constraints on `certification_case_id` in `activity_report_application_forms` and `exemption_application_forms`, which guarantee a single form per case. 

Today, `ReviewActivityReportTask` and `ReviewExemptionClaimTask` do not store a direct reference to the application form they are reviewing. This works because of the above constraints. Before removing the constraints, we want to bind each review task to its specific application form so that resolution no longer depends on the uniqueness invariant. We also want to use this binding in lookup calls.

This ticket does **only** the binding work: it introduces an `application_form` association on each review task class, populates it on task creation and via backfill, and flips the existing read sites to use it. The uniqueness constraints stay in place, so user-visible behavior does not change. The follow-up ticket can then drop uniqueness and add the re-submission flow as a pure behavior change with no data-layer risk left to absorb.

## User Story

As a developer, I want each review task to hold an explicit reference to its application form so that resolving "which form does this task review" does not depend on the form-per-case uniqueness constraint, unblocking the future re-submission work.

### Data model
- [ ] Add an `application_form_id` (UUID) column to `strata_tasks`, with an index
- [ ] On `ReviewActivityReportTask`, declare `belongs_to :application_form, class_name: "ActivityReportApplicationForm", optional: true`
- [ ] On `ReviewExemptionClaimTask`, declare `belongs_to :application_form, class_name: "ExemptionApplicationForm", optional: true`
- [ ] `optional: true` only for the duration of the rollout; tighten to required once backfill is complete and verified (see Rollout)

### Linking on task creation
- [ ] Add an `after_create` callback on `ReviewActivityReportTask` and `ReviewExemptionClaimTask` (or a shared concern) that uses `self.class.application_form_class.find_by(certification_case_id: case_id)` to populate `application_form`
- [ ] Raise if the lookup returns nil (failing loudly is preferable to silently creating an unlinked task — the BP guarantees a submitted form exists at this point)

### Backfill
- [ ] Write a data migration that backfills `application_form_id` for every existing `ReviewActivityReportTask` and `ReviewExemptionClaimTask` row by running the same lookup
- [ ] The migration must be safe to re-run (idempotent) and must report any task it can't backfill rather than skipping silently

### Flip read sites to the new association
[...]

PR #594, opened 2026-05-26
OSCER-585: add application_form_id to tasks

jdettmannnava:
## Ticket

Enables #585

## Changes

- Migration to add application_form_id to tasks

## Context for reviewers

This is the db change for explicitly linking application forms to tasks. The migration also includes setting the application form on the existing tasks using the same logic used in the controllers.

## Testing

Migrations successfully run forward and back. Local testing showed all tasks successfully bound to application forms.

♻️ Environment destroyed ♻️


reporting-app/db/migrate/20260522182945_add_application_form_id_to_strata_tasks.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class AddApplicationFormIdToStrataTasks < ActiveRecord::Migration[8.0]
+  def up
+    add_column :strata_tasks, :application_form_id, :uuid
```
jeffhorn-nava, line 5: Was expecting an index here, from the task description in #585. Should that come here, or in a later migration? Same question for foreign-key constraints on associated application forms.

reporting-app/db/migrate/20260522182945_add_application_form_id_to_strata_tasks.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    [ ReviewExemptionClaimTask, ReviewActivityReportTask ].each do |klass|
+      klass.find_each do |task|
+        application_form = task.class.application_form_class.find_by(certification_case_id: task.case_id)
```
jeffhorn-nava, line 9: The parent issue explicitly wants the migration to "report any task it can't backfill rather than skipping silently". 

`ExemptionApplicationForm` validates `certification_case_id` for uniqueness but not presence (and the schema allows `nil` `case_id` on `strata_tasks`), so a `find_by(certification_case_id: nil)` could match an arbitrary form (that is, whichever is the first one PostgreSQL finds).

Maybe here we could `next if task.case_id.nil?` and use `find_by!` so unmatched rows raise loudly.

reporting-app/db/migrate/20260522182945_add_application_form_id_to_strata_tasks.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class AddApplicationFormIdToStrataTasks < ActiveRecord::Migration[8.0]
+  def up
+    add_column :strata_tasks, :application_form_id, :uuid
+
+    [ ReviewExemptionClaimTask, ReviewActivityReportTask ].each do |klass|
```
jeffhorn-nava, line 7: @giverm discussed some potential gotchas downstream implementers of OSCER might run into. No other migrations in OSCER follow the shadow-class pattern for models, but it's one I've used and recommended in the past.

```
class AddApplicationFormIdToStrataTasks < ActiveRecord::Migration[8.0]
  class StrataTask < ActiveRecord::Base
    self.table_name = "strata_tasks"
  end
  class ExemptionApplicationForm < ActiveRecord::Base; end
  ...
end
```

The idea is to freeze the models to work with this migration as-written, regardless of future drift. My go-to example as a jumping off point: https://www.gregnavis.com/articles/how-to-access-production-data-in-rails-migrations.html

Raising this more for awareness (and to solicit your own and Mike's takes on it) than for a requested change, since this is a new pattern, but one that might serve OSCER going forward. The shadow-class pattern is one of those "correct in principle, ugly in practice" Rails conventions. You see it religiously in codebases that have been burned by migration-replay failures, and ignored in codebases that haven't.

reporting-app/db/migrate/20260522182945_add_application_form_id_to_strata_tasks.rb
jdettmannnava: We aren't going to be looking up task by application form, so no index is necessary

reporting-app/db/migrate/20260522182945_add_application_form_id_to_strata_tasks.rb
jdettmannnava: Yeah, I wrote the ticket based on claude, and I don't really agree with it. I should have edited it out :P

jeffhorn-nava (approved):

reporting-app/db/migrate/20260522182945_add_application_form_id_to_strata_tasks.rb
jdettmannnava: Interesting Idea, but I don't think it would work here. We explicitly only want these two task types to be involved in this migration; I am doing an update_column to avoid any verification work; and I am updating only the field I just created; so I think we will be safe!

jeffhorn-nava: Thanks for sharing your thoughts on your choices here. Makes sense to me!

PR #595, opened 2026-05-27
OSCER-585: Bind application form to review task on task create

### LINKED ISSUE #585: Verification Window: Bind review tasks directly to their application form

jdettmannnava:
## Ticket

Resolves #585 

## Changes

- Review tasks belong to application form
- Tasks controller uses application form linked directly to task
- Task service uses application form linked directly to task

## Context for reviewers

Currently, we use uniqueness of application form type per certification case to bind application forms to review tasks. However, we need to remove this uniqueness constraint. We therefore need to bind the application forms to their review tasks more explicitly.

Creation of review tasks is part of the certification case business process, which supplies only the certification case when creating the task. I am therefore setting the application form based on a left-join lookup before the validation phase. This feels a bit fragile, and it is hard to test as the uniqueness constraint still exists (removing the constraint will be addressed in a later PR), but I think it will be fine in practice. It is at least as robust as the current implementation.

I chose to disable strict loading for the application form, as some of the calls for task are in the strata-sdk-rails, I don't think all requests for review tasks need the application form, and I don't think the performance gains of strict loading in this case are worth "fixing" this.

I also chose to have the base factory for the review tasks not include the form and have most specs use a review task "with form", just because it seems more idiomatic to me.

For testing, this is a refactor, so testing should just be "still works the same".

## Testing

Automated tests all work (pending e2e test fix).
Manually tested exemption and activity-report flow.

♻️ Environment destroyed ♻️


reporting-app/app/models/oscer_task.rb
```diff
@@ -7,4 +7,14 @@ class OscerTask < Strata::Task
   def self.policy_class
     Strata::TaskPolicy
   end
+
+  private
+
+  def ensure_application_form
+    if application_form.nil?
```
jgavin-nava, line 14: Do we need to log anything if there is no application form is found?

reporting-app/app/models/oscer_task.rb
```diff
@@ -7,4 +7,14 @@ class OscerTask < Strata::Task
...
+      application_forms = self.class.application_form_class.joins("LEFT JOIN strata_tasks ON #{self.class.application_form_class.table_name}.id = strata_tasks.application_form_id")
+                                 .where("certification_case_id = ? AND strata_tasks.application_form_id IS NULL", case_id)
+      self.application_form = application_forms.first
```
jgavin-nava, line 18: Should we order this first by created_at so its deterministic

reporting-app/app/models/oscer_task.rb
jdettmannnava: I don't think so, as it will fail validation, which will log as an error

jgavin-nava: Yea good point just wanted to bring it up

jgavin-nava (approved):

PR #612, opened 2026-06-01
OSCER-609 Aggregators use explicit application forms

### LINKED ISSUE #609: Refactor aggregators for multiple-forms per case
## Background

The hours/income compliance services and `ActivityAggregator` concern currently accept a `certification_case:` keyword argument that they immediately use to look up the case's activity-report form via `find_by(certification_case_id:)`. The case was always a stand-in for the form — the services don't actually need the case for anything else.

This indirection works today because of the form-per-case uniqueness constraint: there's exactly one form to find. Once that constraint is removed (next ticket), the implicit `case → form` lookup becomes ambiguous, and every caller would silently fall back to "whichever form the service picks." Better to make the form explicit *now* — under the current uniqueness invariant, where the refactor is a pure no-op — so each callsite carries the form-selection decision when uniqueness drops.

The `certification` positional argument stays. It's doing real work: providing `member_id` for external-source data (which is member-scoped, not form-scoped) and `certification_requirements.continuous_lookback_period` for the income window.

## User Story

As a developer, I want the determination services to accept the application form they should aggregate against, so that the "which form?" decision lives at each callsite — making it explicit, reviewable, and ready for the multi-form world the next ticket introduces.

## Acceptance Criteria

- [ ] The public API of `HoursComplianceDeterminationService` and `IncomeComplianceDeterminationService` accepts the activity-report form directly (in place of the current `certification_case:` keyword) on:
  - `aggregate_hours_for_certification`
  - `aggregate_income_for_certification`
  - `member_hour_activities_for_certification`
  - `member_income_activities_for_certification`
- [ ] `ActivityAggregator#fetch_member_activities` (and any related internal helpers) work directly from the form, not by walking the case.
- [ ] All current callsites are updated to pass the form they intend to aggregate against. Under existing uniqueness, this is mechanical (`case.activity_report_application_form` or equivalent).
- [ ] The `certification_case_for_certification` tie-break helper in `ActivityAggregator` is removed if it has no remaining consumers (it exists today to disambiguate the multiple-cases-per-certification edge case that the case-based lookup created).
- [ ] No behavior change: the existing test suite passes without weakening assertions, and any test that previously passed a case now passes the corresponding form.

## Out-of-scope

- Dropping the form-per-case uniqueness constraints — that's the next ticket; this one lands cleanly under the current invariant.
- Resolving the "which form?" decision for the multi-form world — each callsite simply passes the single form that exists today. The real decisions land in the uniqueness-drop ticket.
- **`.calculate(certification_id)` entry points stay as-is — do not delete or refactor.** `HoursComplianceDeterminationService.calculate` currently has zero callers and `IncomeComplianceDeterminationService.calculate` has exactly one (`ExternalIncomeActivityService`). Both are *intentional future-proofing* for the scenario where external data arriving during the application lifecycle silently resolves a case without a caller having a form in hand. They continue to take only `certification_id` and resolve the case/form internally. Their multi-form semantics belong to a follow-up ticket.
- Member-facing UI changes.

## Dependencies & Risks

- **Pure refactor under current uniqueness.** Tests are the safety net — under the one-form-per-case invariant, this should be behaviorally equivalent. If a test changes shape, that's a signal worth investigating.
- **Sequencing matters.** This ticket must land *before* the uniqueness-drop ticket. If they ship in the wrong order, the uniqueness-drop ticket has to keep the "expedient stopgap" language for the compliance services.
- **Callsites in scope:** `CertificationCase` model methods (recording determinations), `CertificationCasesController` (staff case page), `DashboardController` / `MemberDashboardCompliance(Service)` (member dashboard), `NotificationsEventListener` (adverse-decision email recomputes), `CommunityEngagementCheckService` (periodic check), `ExternalIncomeActivityService` (external-data path through `.calculate`). About 8 callsites total.

## Links & Resources

- Services to refactor: `app/services/hours_compliance_determination_service.rb`, `app/services/income_compliance_determination_service.rb`, `app/services/concerns/activity_aggregator.rb`
- Callsites: `app/models/certification_case.rb:52,74`, `app/controllers/certification_cases_controller.rb:29,34,41,46`, `app/services/member_dashboard_compliance_service.rb:25`, `app/models/member_dashboard_compliance.rb:148,151`, `app/services/notifications_event_listener.rb:49,72,83`, `app/services/community_engagement_check_service.rb:13,17`, `app/services/external_income_activity_service.rb:59`
- Follow-up that this enables: the drop-uniqueness ticket (`reporting-app/tmp/allow-form-resubmission-within-case.md`)

jdettmannnava:
## Ticket

Resolves #609

## Changes

- Hours and Income determination signatures ask for activity report application forms instead of the certification case.

## Context for reviewers

This is just a refactor.

Hours and income determinations used to grab the activity report application forms from the certification case in order to compile data. As we move to multiple activity report application forms, we need to make the form explicit.

Some places (e.g. dashboard controller) do not have a clear context for which application form is in play. Future tickets will address behavior in these cases.

## Testing

- All automated tests pass
- Manually checked accept and deny activity reports
 - Member Dashboard
 - Staff Certification Case
 - Staff Tasks
 - Determination
 - Notification (deny)

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/services/hours_compliance_determination_service.rb
+++ b/reporting-app/app/services/hours_compliance_determination_service.rb
@@ -6,12 +6,6 @@ class HoursComplianceDeterminationService
   class << self
     include ActivityAggregator
 
-    # Called by CertificationBusinessProcess after ActivityReportApproved event
-    # @param kase [CertificationCase]
-    def determine_after_activity_report(kase)
-      perform_determination(kase, not_compliant_event: "DeterminedHoursInsufficient")
-    end
-
     # Called by CalculateComplianceJob for async recalculation of existing certifications
     # Records determination without triggering workflow events/notifications.
     # When compliant, +record_hours_compliance+ closes the case (+CertificationCase#record_automated_ce_compliance+);
@@ -89,30 +83,6 @@ class HoursComplianceDeterminationService
     end
 
     private
-
-    # Shared logic for both initial and post-activity-report determination
-    # @param kase [CertificationCase]
-    # @param not_compliant_event [String] event name to publish when not compliant
-    def perform_determination(kase, not_compliant_event:)
-      certification = Certification.find(kase.certification_id)
-      hours_data = aggregate_hours_for_certification(certification)
-      outcome = determine_outcome(hours_data[:total_hours])
-
-      kase.record_hours_compliance(outcome, hours_data)
-
-      if outcome == :compliant
-        Strata::EventManager.publish("DeterminedHoursMet", {
-          case_id: kase.id,
-          certification_id: certification.id
-        })
-      else
-        Strata::EventManager.publish(not_compliant_event, {
-          case_id: kase.id,
-          certification_id: certification.id,
-          hours_data: hours_data
-        })
-      end
-    end
 
     def determine_outcome(total_hours)
       compliant_for_total_hours?(total_hours) ? :compliant : :not_compliant
--- a/reporting-app/app/services/hours_compliance_determination_service.rb
+++ b/reporting-app/app/services/hours_compliance_determination_service.rb
@@ -39,7 +39,7 @@ class HoursComplianceDeterminationService
     # @param external_hourly_activities [ActiveRecord::Relation<ExternalHourlyActivity>, Array<ExternalHourlyActivity>, nil]
     #   When set, skips fetching external rows again (e.g. staff +#show+ already loaded them).
     # @param member_hour_activity_rows [Array<Activity>, nil] When set, skips
-    #   +member_hour_activities_for_certification+ for totals/ids (rows must match +certification_case:+ when passed).
+    #   +member_hour_activities_for_certification+ for totals/ids (rows must match +application_form:+ when passed).
     # @return [Hash] with total_hours, hours_by_category, hours_by_source, etc.
     def aggregate_hours_for_certification(
       certification,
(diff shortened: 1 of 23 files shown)


reporting-app/app/services/notifications_event_listener.rb
```diff
@@ -81,8 +80,7 @@ def handle_insufficient_community_engagement(event)
           payload[:income_data]
         else
           IncomeComplianceDeterminationService.aggregate_income_for_certification(
-            certification,
-            certification_case: certification_case_for_notification(certification, payload)
```
giverm, line 85: `certification_case_for_notification` is now unused

reporting-app/app/services/hours_compliance_determination_service.rb
```diff
@@ -74,17 +73,12 @@ def aggregate_hours_for_certification(
...
     # @param certification [Certification]
-    # @param certification_case [CertificationCase, nil] When set, uses this case's activity report. When nil,
-    #   uses +ActivityAggregator#certification_case_for_certification+ (deterministic tie-break) to resolve.
+    # @param application_form [ActivityReportApplicationForm, nil]
     # @return [ActiveRecord::Relation<Activity>]
-    def member_hour_activities_for_certification(certification, certification_case: nil)
-      kase = certification_case_for_certification(certification, certification_case)
-      return Activity.none unless kase
+    def member_hour_activities_for_certification(certification, application_form:)
+      return Activity.none unless application_form
```
giverm, line 73: Tracing through the calls in this service, this return will cause the `aggregate_hours_for_certification` call in `perform_determination` to always return 0. Luckily `perform_determination` is only called by `determine_after_activity_report`, which appears to be unused. I think dropping those two methods, either here or a separate PR, should resolve any concerns.

giverm (approved):
Looks good! Just some dead code cleanup

PR #619, opened 2026-06-02
OSCER-608 allow case to have multiple application forms of each type

### LINKED ISSUE #608: Remove uniqueness constraint on application forms
## Background

A `CertificationCase` is currently limited to a single `ActivityReportApplicationForm` and a single `ExemptionApplicationForm`, enforced by `uniqueness: true` validations on both models and unique DB indices on `certification_case_id`. Members cannot re-submit a form within the same case after a denial.

The prerequisite plumbing is already in place: #585 bound review tasks directly to their `application_form` via a polymorphic association, so resolving "which form does this task review" no longer depends on the form-per-case uniqueness invariant. This story drops the uniqueness restriction so that a case can hold multiple forms of the same type over its lifecycle (one per submission cycle), and updates read sites that currently assume singularity to resolve a sensible single form for their context.

Reviews remain **sequential by design** — the business process guarantees that at any moment, at most one form of a given type on a case is awaiting review. Concurrent in-flight reviews are not introduced here.

## User Story

As a member whose activity report or exemption claim has been denied, I want to submit a new form within the same certification case so that I can act on the denial without losing the continuity of my case.

## Acceptance Criteria

- [ ] A `CertificationCase` can hold multiple `ActivityReportApplicationForm` records and multiple `ExemptionApplicationForm` records over its lifecycle.
- [ ] Model-level and DB-level uniqueness constraints on `certification_case_id` are removed from both form models.
- [ ] **Member dashboard**: resolves the active (non-terminal) form for the case if one exists; otherwise the most recent form.
- [ ] **Staff certification case page**: resolves the most convenient form for the case (no list of historical forms yet — that's a separate ticket).
- [ ] **Compliance / aggregator services** (`HoursComplianceDeterminationService`, `IncomeComplianceDeterminationService`, `ActivityAggregator`): keep working without raising or returning incorrect results in the multi-form world. 
- [ ] Review-task creation continues to bind to the correct submitted form on the case (the #585 mechanism still works when more than one form exists historically — the "unlinked submitted form" lookup remains unambiguous because reviews are sequential).
- [ ] Existing single-form cases behave exactly as they do today.

## Out-of-scope

- Member-facing UI / flow for re-submitting a form after a denial — separate story. This ticket unblocks that work at the data layer; the user-visible re-submission experience is its own behavior change.
- Proper semantics for compliance / aggregator services in the multi-form world — separate ticket. The expedient implementation here is a stopgap.
- Staff UI for viewing the full history of a case's forms (current + past) — separate ticket.
- Concurrent in-flight reviews of multiple forms of the same type on a case — reviews stay sequential.
- Backfill of historical data — no historical data needs migrating; the constraint simply stops being enforced going forward.

## Dependencies & Risks

- **Sequentiality must hold.** The task→form binding installed by OSCER-585 finds the most recently created, unlinked submitted form on the case at task-creation time. That lookup is unambiguous only because the business process never creates a new review task while another review of the same form type is in flight. Any future change that could create two unlinked submitted forms of the same type at once would break the binding.
- **Compliance services are a known weak spot.** Their multi-form semantics are deferred — be alert that determinations on cases with multiple forms could be misleading until the follow-up lands. Worth a code comment at the resolution site flagging the deferral.
- **No data migration required**, but the unique index drop should be a routine migration; confirm it runs without locking issues on the affected tables.

## Links & Resources

- Uniqueness to remove: `app/models/activity_report_application_form.rb:15`, `app/models/exemption_application_form.rb:9`, plus the unique indices on both `*_application_forms.certification_case_id` in `db/schema.rb`
- Read sites that assume singularity: `app/controllers/certification_cases_controller.rb` (~lines 22, 78–80), `app/controllers/dashboard_controller.rb` (~lines 31, 35), `app/services/hours_compliance_determination_service.rb`, `app/services/income_compliance_determination_service.rb`, `app/services/concerns/activity_aggregator.rb`
- Task→form binding (already in place, this story relies on it): `app/models/oscer_task.rb:13-20`, `app/models/review_activity_report_task.rb`, `app/models/review_exemption_claim_task.rb`

jdettmannnava:
## Ticket

Resolves #608 

## Changes

- Change uniqueness on application forms to uniqueness on in_progress forms
- Update member dashboard to return in_progress form(s) if exists, else arbitrary form(s)

## Context for reviewers

Final step in the journey in the backend to allow multiple form submissions in the verification window.

Does *not* enable users to submit multiple forms.

FWIW, the dashboard test failed without the "order by status" set when testing locally.

Manually tested only for regressions, as the frontend does not even support submitting an activity report application form when an exemption form has been denied.

## Testing

- CI passes
- Manual test of member dashboard to assure no regressions
  - Works when go back to dashboard with in-progress activity report form
  - Works when go to dashboard with submitted activity report form
  - Works when go back to dashboard with in-progress exemption form
  - Works when go to dashboard with submitted exemption form

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/controllers/dashboard_controller.rb
+++ b/reporting-app/app/controllers/dashboard_controller.rb
@@ -28,11 +28,11 @@ class DashboardController < ApplicationController
   end
 
   def set_exemption_application_form
-    @exemption_application_form = ExemptionApplicationForm.where(certification_case_id: @certification_case&.id).order(:status).first
+    @exemption_application_form = ExemptionApplicationForm.where(certification_case_id: @certification_case&.id).order(:status, created_at: :desc).first
   end
 
   def set_activity_report_application_form
-    @activity_report_application_form = ActivityReportApplicationForm.where(certification_case_id: @certification_case&.id).order(:status).first
+    @activity_report_application_form = ActivityReportApplicationForm.where(certification_case_id: @certification_case&.id).order(:status, created_at: :desc).first
   end
 
   def set_information_requests
--- a/reporting-app/app/controllers/dashboard_controller.rb
+++ b/reporting-app/app/controllers/dashboard_controller.rb
@@ -28,11 +28,11 @@ class DashboardController < ApplicationController
   end
 
   def set_exemption_application_form
-    @exemption_application_form = ExemptionApplicationForm.find_by_certification_case_id(@certification_case&.id)
+    @exemption_application_form = ExemptionApplicationForm.where(certification_case_id: @certification_case&.id).order(:status).first
   end
 
   def set_activity_report_application_form
-    @activity_report_application_form = ActivityReportApplicationForm.find_by_certification_case_id(@certification_case&.id)
+    @activity_report_application_form = ActivityReportApplicationForm.where(certification_case_id: @certification_case&.id).order(:status).first
   end
 
   def set_information_requests
--- a/reporting-app/app/models/activity_report_application_form.rb
+++ b/reporting-app/app/models/activity_report_application_form.rb
@@ -12,7 +12,8 @@ class ActivityReportApplicationForm < Strata::ApplicationForm
   strata_attribute :number_of_months_to_certify, :integer
   strata_attribute :months_that_can_be_certified, :year_month, array: true
 
-  validates :certification_case_id, presence: true, uniqueness: true
+  # At most one in-progress form per case; any number of submitted forms may accumulate over resubmissions.
+  validates :certification_case_id, presence: true, uniqueness: { conditions: -> { in_progress } }
   validates :reporting_periods,
     length: {
       is: :number_of_months_to_certify,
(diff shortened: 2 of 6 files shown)


reporting-app/app/controllers/dashboard_controller.rb
```diff
@@ -28,11 +28,11 @@ def set_certification_case
   end
 
   def set_exemption_application_form
-    @exemption_application_form = ExemptionApplicationForm.find_by_certification_case_id(@certification_case&.id)
+    @exemption_application_form = ExemptionApplicationForm.where(certification_case_id: @certification_case&.id).order(:status).first
```
giverm: From the AC:
> Member dashboard: resolves the active (non-terminal) form for the case if one exists; otherwise the most recent form.

This query will resolve the active form, but if there is no active form it isn't guaranteed to return the most recent if there are multiple submitted forms. 

```suggestion
    @exemption_application_form = ExemptionApplicationForm.where(certification_case_id: @certification_case&.id).order(:status, created_at: :desc).first
```

A couple other thoughts:
- Since the allowed status values are defined in the SDK, I had to go on a little journey to understand the sort. Might be worth a comment explaining the sort, but totally up to you. 
- Would it be worth adding a model method for this query?

giverm (approved):

reporting-app/app/controllers/dashboard_controller.rb
jdettmannnava: Fixed it and added a test!

PR #627, opened 2026-06-03
OSCER-410: Log external income entries

jdettmannnava:
## Ticket

Resolves #410

## Changes

- Audit log line created on ExternalIncomeActivity creation
- Raise ActiveRecord validation errors in ExternalIncomeActivityService instead of downstream.

## Context for reviewers

Instead of strict immutability, we opted for logging. The payload is based on the ticket.

Currently, on failure to save the ExternalIncomeActivity, ExternalIncomeActivityService returns the validation errors in a hash, which are later used to raise an ActiveRecord validation error. This implementation feels cleaner for this use case.

## Testing

- CI
- Validated log from demo service creation with partial income met external scenario

♻️ Environment destroyed ♻️


reporting-app/app/services/external_income_activity_service.rb
```diff
@@ -39,11 +40,19 @@ def create_entry(member_id:, category:, gross_income:, period_start:, period_end
...
-        { error: entry.errors.full_messages.join(", ") }
+      ActiveRecord::Base.transaction do
+        if entry.save
+          Strata::AuditLog.write!(
+            actor: self,
+            action: "external_income_activity.create",
+            subject: entry,
+            data: entry.attributes
+          )
+          maybe_recalculate_income_compliance(entry.member_id) if recalculate_income_compliance
```
jgavin-nava: This might have just been an oversight before or not thought about with this change it brings up this questions:

Was the transaction meant to cover compliance too, or only save + audit? Wrapping maybe_recalculate_income_compliance means a compliance failure rolls back the income row (pre-PR it would stay persisted). Fine if intentional — suggest a test + PR note. If not, maybe we just limit the transaction to entry.save + Strata::AuditLog.write! and running compliance after.

jdettmannnava: Nope, hadn't noticed

reporting-app/app/services/external_income_activity_service.rb
jgavin-nava: Only thing I might change here is the documentation


```suggestion
    # @return [Hash] with +:error+ key when a duplicate entry is detected (before save)
    # @raise [ActiveRecord::RecordInvalid] when validations fail
```

jgavin-nava (approved):

Issue #480, opened 2026-04-20
Implement UI for income member compliance progress


malworks:
## Background

The member dashboard needs to display income progress and compliance status so members can understand where they stand against the community engagement requirement at a glance. This ticket covers implementing the UI designed in [OSCER-337](https://www.figma.com/design/GeDX8dYzAHytqyp1v1stuZ/-OSCER-337--Member-income-progress-UI---compliance-status-display)

## User Story

As a member, I want to see my income reporting progress and current compliance status on my dashboard so that I know how much more I need to report and whether my Medicaid coverage is at risk.

## Acceptance Criteria

- UI should match the Figma designs
- Exempt status badges are displayed according to the logic found in the legend
- Progress bar in the "Income reported" card reflects the percentage of the $580 requirement met, and updates based on proximity to the deadline/requirement
- Report status card shows one of the following states with appropriate visual treatment: `In progress`, `Under review`, `Compliant`, `Not compliant`

## Links & Resources

- [Figma – OSCER-337 Member income progress UI / compliance status display](https://www.figma.com/design/GeDX8dYzAHytqyp1v1stuZ/-OSCER-337--Member-income-progress-UI---compliance-status-display?node-id=1-3)
- Design ticket: #337

jgavin-nava:
## Split into stacked PRs (OSCER-337 / Figma)

Parent epic — implementation split into five child issues for reviewable stacked PRs:

| Order | Issue | Scope |
|------:|-------|-------|
| 1 | #639 | Landing page (welcome hero, get started) |
| 2 | #640 | Exemption outcomes (exempt, under review, does not qualify) |
| 3 | #641 | Exemption draft in progress |
| 4 | #642 | Reporting started — **activity tables only** (staff-style: income and/or hours tables based on reported data) |
| 5 | #643 | Progress cards + report status card (original AC for bars/status) |

**Stack order:** merge #639 → #640/#641 → #642 → #643.

PR #626 will be retargeted or replaced by this sequence.

Vault: `nava-notes/Tickets/OSCER-480-split-plan.md`

PR #646, opened 2026-06-08
OSCER-637: Allow multiple exemption forms

### LINKED ISSUE #637: Allow multiple exemption application forms
## Background
Members need to be able to submit new exemption application forms as long as the verification window has not closed.

## User Story

As a member I want to be able to submit new exemption application forms so that I can exercise my rights within the verification window.

## Acceptance Criteria

- [ ] Member can submit new exemption application forms as long as within verification window
- [ ] Member cannot submit new exemption application forms when case closed
- [ ] Member cannot submit new exemption application forms when verification window closed

## Out-of-scope
Changes to exemption application form wizard are out of scope 

## Open questions 
Does the verification window need to be open to resubmit?

[Tech spec for verification window](https://docs.google.com/document/d/10raEsJU-dEzJ4ExXzcWQXpHWAailVh5UoUm73IIBSAk/edit?tab=t.0)

jdettmannnava:
## Ticket

Resolves #637

## Changes

- Changed dashboard helper to use new `flow_status` method on exemption application form
- Added "submit new form" button to exemption-denied partial
- Added `has_pending_form` method to exemption application form class
- Block new exemption application form when existing exemption form in flight, verification window ended, or case closed
- Error handling when case closed and try to make new exemption application form
- Added `shell` option to Makefile

## Context for reviewers

This feature allows a member to submit another ExemptionApplicationForm if a previous form was denied. Allowing multiple ActivityReportApplicationForms is another ticket.

There seems to be some issues on which form is shown in the 'view form' button, but the UI for these changes is being updated in #480, so fixing it here feels low-priority.

The `shell` option was added to run specific specs more easily.

## Testing

- Submitted multiple forms (denying each time)
- Manually closed case
- Manually returned 'true' for existing-form check

new
closed
pending-form

♻️ Environment destroyed ♻️


reporting-app/app/models/exemption_application_form.rb
```diff
@@ -26,4 +27,34 @@ def event_payload
...
+                                                   status: :completed).exists?
+    return status unless task_complete
+
+    CertificationCase.find(certification_case_id).exemption_request_approval_status
+  end
+
+  def self.has_pending_form(certification_case_id)
+    ExemptionApplicationForm.where(certification_case_id:, status: :in_progress).exists? ||
+    ReviewExemptionClaimTask.where(application_form: ExemptionApplicationForm.where(certification_case_id:).all,
+                                   status: :pending).exists?
```
jgavin-nava: I believe this should also include the status `on_hold` for Tasks. As a caseworker could send an RFI which I bleieve puts that task into an on_hold status and then they could submit again.

reporting-app/app/models/exemption_application_form.rb
```diff
@@ -26,4 +27,34 @@ def event_payload
...
+  end
+
+  private
+
+  def case_not_closed
+    certification_case = CertificationCase.find_by(id: certification_case_id)
+    if certification_case.closed?
+      errors.add(:certification_case_id, "has closed")
+    elsif certification_case.verification_window_end_date && certification_case.verification_window_end_date < Time.now
```
jgavin-nava, line 60: I think the logic around this gets messed up because `verification_window_end_date` is a Date and comparing it with `Time.now` I believe makes it seem like the verifciation window is closed a day early since the date is coerced to midnight

reporting-app/app/models/exemption_application_form.rb
```diff
@@ -26,4 +27,34 @@ def event_payload
...
+    ExemptionApplicationForm.where(certification_case_id:, status: :in_progress).exists? ||
+    ReviewExemptionClaimTask.where(application_form: ExemptionApplicationForm.where(certification_case_id:).all,
+                                   status: :pending).exists?
+  end
+
+  private
+
+  def case_not_closed
+    certification_case = CertificationCase.find_by(id: certification_case_id)
+    if certification_case.closed?
```
jgavin-nava: This could cause a validation error if there is an issue with the certification case. 


```suggestion
    return errors.add(:certification_case_id, "is invalid") if certification_case.blank?

    if certification_case.closed?
```

reporting-app/app/controllers/exemption_application_forms_controller.rb
```diff
@@ -122,6 +122,8 @@ def create_exemption_application_form
     rescue ActiveRecord::RecordInvalid => e
       if e.record.errors[:certification_case_id].include?("has already been taken")
         redirect_to dashboard_path, notice: "An exemption application already exists for this certification case"
+      elsif e.record.errors[:certification_case_id].include?("has closed")
+        redirect_to dashboard_path, notice: "Certification case has been closed"
       else
```
jgavin-nava, line 129: Should we add one for the verification window being closed as well?

reporting-app/app/controllers/exemption_screener_controller.rb
jgavin-nava, line 143: Should we handle creation errors here as well. For case closed and verification window?

reporting-app/app/views/dashboard/_exemption_denied.html.erb
```diff
@@ -12,5 +12,8 @@
             class: "usa-button margin-right-2 margin-bottom-1" %>
     <% end %>
     <%= link_to t(".view_exemption_button"), exemption_application_form_path(exemption_application), class: "usa-button usa-button--outline margin-bottom-1" %>
+    <%= link_to(t(".submit_new_exemption_button"),
+	new_exemption_application_form_path(certification_case_id: certification_case&.id),
+	class: "usa-button usa-button--outline margin-bottom-1") unless certification_case&.closed? || ExemptionApplicationForm.has_pending_form(certification_case&.id) %>
```
jgavin-nava: Should there also be a check for the verification window not being open or would that be closed already since the window is closed?

reporting-app/app/controllers/exemption_application_forms_controller.rb
```diff
@@ -122,6 +122,8 @@ def create_exemption_application_form
     rescue ActiveRecord::RecordInvalid => e
       if e.record.errors[:certification_case_id].include?("has already been taken")
         redirect_to dashboard_path, notice: "An exemption application already exists for this certification case"
+      elsif e.record.errors[:certification_case_id].include?("has closed")
+        redirect_to dashboard_path, notice: "Certification case has been closed"
```
jgavin-nava: I know the one above this doesnt have it either but should we move these hardcoded strings to a locale?

reporting-app/app/models/exemption_application_form.rb
```diff
@@ -26,4 +27,34 @@ def event_payload
   def self.information_request_class
     ExemptionInformationRequest
   end
+
+  def flow_status
```
jgavin-nava, line 31: I think this gets called multiple times when the dashboard helper runs so if we could use memoizing and reusing the already loaded case that would probably help performance

reporting-app/app/controllers/exemption_screener_controller.rb
jdettmannnava: Adding another before action.

reporting-app/app/models/exemption_application_form.rb
jdettmannnava: I think the end date is the close date? The 30-day period starts the day of communication, so that day + 30 should be closed.

jgavin-nava: I understood end date as the last that it can be verified. But if it is the first day it is closed then we are good

jgavin-nava (approved):

jgavin-nava:
LGTM

PR #651, opened 2026-06-10
Add database migration convention for downstream replay safety

### LINKED ISSUE #543: Open Source Update Strategy | Phase 2 | Keep migrations simple and separable (audit + convention)
## User story

As a state implementer running `nava-platform app update . reporting-app` on a deployment with customizations or a large production dataset, I want OSCER's migrations to follow a predictable convention so that I can apply them in a controlled order and recover if one fails mid-way. The convention should keep schema and data changes in separate files, prefer reversible migrations, and avoid references to app-level model classes that may not exist when a deployment replays the migration years later.

## Context

OSCER is moving toward a release model where downstream deployments run `nava-platform app update . reporting-app` to pull new versions. That update step ends with `bin/rails db:migrate` against the deployment's own production database. Today's `db/migrate/` directory (52 migrations as of 2026-05-06) wasn't written with that constraint in mind — several migrations couple schema and data changes in ways that work on Nava's environments but create avoidable failure modes for an external deployment replaying them later against a different dataset.

A quick audit surfaces several patterns the convention needs to address. Concrete examples:

- **Mixed schema + data in one migration** (DDL and data backfill in the same `up`):
  - `20251021141137_convert_activity_to_sti.rb` — `add_column :activities, :type` followed by `Activity.where(type: nil).update_all(type: 'WorkActivity')` in the same migration
  - `20251031214909_update_determination_reasons.rb` — adds a new array column, backfills it via raw SQL, removes the old column, all in one `up`
  - `20260305120000_add_uploader_constraint_by_source_type.rb` — data `UPDATE` to satisfy a CHECK constraint, then `ALTER TABLE ADD CONSTRAINT`, in the same `up`

- **Migration depends on app-code state at replay time:**
  - `20251021141137_convert_activity_to_sti.rb` references the `Activity` model constant inside the migration. If `Activity` is later renamed or removed, replaying this migration on a fresh deployment install will fail at the constant lookup. The standard Rails workaround is to use raw SQL or an anonymous AR class scoped to the migration (`Class.new(ActiveRecord::Base) { self.table_name = 'activities' }`).

- **Counter-example (already follows the desired pattern):**
  - `20251009164541_migrate_reporting_period_to_reporting_periods.rb` — pure data migration, raw SQL only, no model references, with both `up` and `down` defined.

**Effort estimate:** ~1-2 days. The audit is mechanical (grep + read each flagged migration); the convention itself is short. The judgment call is what to do about the existing violations — that decision drives the size of the follow-up work, if any.

## Sequencing

Independent of the other Phase 2 work; no shared files. Most valuable to land before the first external deployment runs an upgrade against a non-Nava database.

## Scope

**In this ticket:**

- **Audit deliverable:** a markdown table in `docs/reporting-app/migrations.md` listing the migrations that violate the proposed convention, what they violate, and a recommended disposition (grandfather as-is / split in a follow-up / rewrite). The audit is the evidence that stress-tests the convention; if the convention is too strict to be practical, the audit will surface that.
- **Convention** committed to a short reference doc, covering at minimum:
  - **Schema vs. data separation:** schema-only migrations (DDL) and data-only migrations live in separate files. Common multi-step patterns (add column → backfill → drop old column) become a sequence of three migrations, not one.
  - **Reversibility:** prefer `change` when Rails can auto-reverse; otherwise both `up` and `down` are defined. Document the exceptions where `down` is genuinely impractical (e.g., destructive data migrations).
  - **No model class references in migrations:** use raw SQL or migration-local anonymous AR classes. Migrations must replay correctly on a fresh deployment install years from now without depending on app-level constants still existing.
  - **Required upgrade stops:** when a data migration assumes a specific intermediate state (e.g., "must upgrade to 2026.2.0 first to run the backfill"), call it out in the release notes for the gating version.
- **Enforcement** decision (Rubocop rule / PR-review checklist entry / convention-doc-only) recorded in the doc.
- **Document the install path vs. update path for database setup.** Fresh-install entry points should use `db:schema:load` + `db:seed` (one-shot from `db/schema.rb`, no migration replay). Update entry points use `db:migrate` (incremental). Audit `Makefile`, deploy scripts, and Dockerfiles to confirm each entry point uses the right command. Today `make db-migrate` correctly uses `db:migrate` for the update path, and `make db-reset` correctly uses `schema.rb` for full recreation — verify there's no fresh-install path that incorrectly invokes `db:migrate` and replays all 52 historical migrations. This narrows the migration-replay-safety audit (the rest of this ticket) to the update path where it actually matters.

**Out of this ticket:**

- Rewriting the existing-violation migrations. The audit identifies them and recommends a disposition; actually splitting `convert_activity_to_sti` (or any other) is a follow-up if the disposition warrants it. This ticket does not block on that follow-up.
- Migration-ordering or zero-downtime guidance (e.g., the `strong_migrations` gem's checks). That's a separate production-safety conversation; the concern here is the upgrade path for downstream deployments.
- Backfilling release notes for past versions. The required-upgrade-stops convention applies forward from the first tagged release.

## Acceptance criteria

- [ ] Audit table in `docs/reporting-app/migrations.md` lists each migration that violates the proposed convention, with the specific violation and a recommended disposition
- [ ] Convention doc covers: schema/data separation, reversibility, no model-class references, required-upgrade-stop documentation
- [ ] Decision recorded on enforcement mechanism (Rubocop rule / PR-review checklist / doc-only)
- [ ] Doc reviewed by team

## Open questions

- **Disposition of existing violations.** Do we (a) grandfather the four flagged migrations as documented exceptions, (b) split them in a follow-up before the first tagged release, or (c) split them lazily as deployments report friction? Recommendation: (a) — they've already run on Nava's environments without incident, and rewriting historical migrations risks creating worse problems than it solves; the convention applies forward.
- **Enforcement mechanism.** `rubocop-rails` (version 2.33 in this repo) ships 9 migration-related cops, all disabled by default. The ones most relevant to this convention:

  - [`Rails/ReversibleMigration`](https://docs.rubocop.org/rubocop-rails/latest/cops_rails.html#railsreversiblemigration) — checks whether the `change` method is reversible (direct fit for the reversibility rule)
  - [`Rails/ReversibleMigrationMethodDefinition`](https://docs.rubocop.org/rubocop-rails/latest/cops_rails.html#railsreversiblemigrationmethoddefinition) — requires either `change`, or both `up` and `down`
  - [`Rails/NotNullColumn`](https://docs.rubocop.org/rubocop-rails/latest/cops_rails.html#railsnotnullcolumn) — prevents adding `NOT NULL` columns without defaults (the unsafe pattern that motivates the schema/data-separation rule)

  Other available migration cops are mostly hygiene: [`Rails/AddColumnIndex`](https://docs.rubocop.org/rubocop-rails/latest/cops_rails.html#railsaddcolumnindex), [`Rails/BulkChangeTable`](https://docs.rubocop.org/rubocop-rails/latest/cops_rails.html#railsbulkchangetable), [`Rails/CreateTableWithTimestamps`](https://docs.rubocop.org/rubocop-rails/latest/cops_rails.html#railscreatetablewithtimestamps), [`Rails/DangerousColumnNames`](https://docs.rubocop.org/rubocop-rails/latest/cops_rails.html#railsdangerouscolumnnames), [`Rails/MigrationClassName`](https://docs.rubocop.org/rubocop-rails/latest/cops_rails.html#railsmigrationclassname), [`Rails/ThreeStateBooleanColumn`](https://docs.rubocop.org/rubocop-rails/latest/cops_rails.html#railsthreestatebooleancolumn).

  **Gaps:** `rubocop-rails` does not ship a cop for schema/data separation in a single migration, nor for forbidding model-class references inside migrations. Those would need either the third-party [`rubocop-migration`](https://github.com/r7kamura/rubocop-migration) gem (17 cops inspired by strong_migrations, also all disabled by default), the [`strong_migrations`](https://github.com/ankane/strong_migrations) gem (runtime checks at migration time rather than lint), or custom cops.

  Recommendation: enable the two `rubocop-rails` reversibility cops in this ticket (low-cost, direct fit on existing tooling). Treat schema/data separation and the no-model-class-references rules as PR-review and convention-doc-only for now; revisit if violations recur.

Part of #527.

giverm:
## Ticket

Closes #543.

## Summary

OSCER is open source: a downstream deployment keeps its own copy and
periodically syncs upstream OSCER changes in, a step that ends in `bin/rails
db:migrate` against the deployment's own production database. That database may
be many versions behind and hold a dataset unlike Nava's, so a migration that
runs cleanly here can still fail on replay elsewhere (a renamed model constant,
a data backfill coupled to a schema change that assumes a particular state).

This adds a written convention for keeping migrations replay-safe, an audit of
the existing migrations against it, and lightweight enforcement on new
migrations.

## Changes

- **`docs/reporting-app/migrations.md`** (new): the convention (schema/data
  separation, prefer reversible migrations, no model-class references inside
  migrations, required-upgrade-stop release notes) plus an audit table of
  pre-convention migrations with their disposition.
- **`.rubocop.yml`**: enables six migration cops that apply forward to new
  migrations. Two cover the reversibility rule
  (`Rails/ReversibleMigration`, `Rails/ReversibleMigrationMethodDefinition`);
  four cover migration safety and hygiene with no pre-existing violations
  (`Rails/NotNullColumn`, `Rails/AddColumnIndex`, `Rails/DangerousColumnNames`,
  `Rails/MigrationClassName`).
- **`.rubocop_todo.yml`** (new): grandfathers three pre-convention migrations
  that trip `Rails/ReversibleMigration`. These are documented exceptions, not a
  burn-down backlog: they have already run on Nava environments without
  incident, and the convention applies forward only.

## Notes

The enforcement decision (the ticket's open question) landed as: enable the
two reversibility cops plus four zero-violation hygiene cops now, and keep the
schema/data-separation and no-model-references rules as convention-doc and
PR-review guidance rather than custom cops, since `rubocop-rails` ships no cop
for either and recurrence has not been observed. Existing violations are
grandfathered rather than rewritten, matching the ticket's recommendation.

## Testing

Documentation and lint-config only; no application code changed. `make lint`
passes against the new cop set.

♻️ Environment destroyed ♻️

### DIFF
--- a/docs/reporting-app/migrations.md
+++ b/docs/reporting-app/migrations.md
@@ -1,0 +1,221 @@
+# Database migrations
+
+OSCER is open source. A downstream deployment keeps its own copy and
+periodically syncs upstream OSCER changes in, a process that ends with `bin/rails
+db:migrate` against the deployment's own production database, which may be years
+of versions behind and hold a dataset that looks nothing like Nava's.
+
+That replay model is the reason this convention exists. A migration that works
+fine on Nava's environments can still fail when an external deployment replays
+it later: it might reference a model class that has since been renamed, or
+couple a schema change to a data backfill that assumes a particular
+intermediate state. The rules below keep migrations replayable and recoverable
+for downstream deployments.
+
+## The convention
+
+These rules apply to **new** migrations. Migrations that predate the convention
+are grandfathered (see [Audit of existing
+migrations](#audit-of-existing-migrations)); they are not a backlog to rewrite.
+
+### 1. Separate schema changes from data changes
+
+Schema-only migrations (DDL) and data-only migrations (DML) live in separate
+files. A common multi-step change such as "add a column, backfill it, drop the
+old column" becomes a sequence of separate migrations, not one:
+
+```
+20260101120000_add_reasons_to_determinations.rb       # DDL: add_column :reasons
+20260101120001_backfill_determination_reasons.rb      # DML: copy reason -> reasons
+20260101120002_remove_reason_from_determinations.rb   # DDL: remove_column :reason
+```
+
+Separation gives a deployment a controlled order and a recovery point: if the
+backfill fails against an unexpected dataset, the schema change has already
+committed cleanly and the data step can be re-run or fixed in isolation, rather
+than rolling back a half-applied combined migration.
+
+### 2. Prefer reversible migrations
+
+Use `change` when Rails can auto-reverse the operation. When it cannot, define
+both `up` and `down`. Two common operations are **not** auto-reversible and
+need explicit help:
+
+- `drop_table` needs a block describing the table so it can be recreated on rollback.
+- `remove_column` needs the column type (`remove_column :table, :col, :string`) so it can be re-added.
+
+```ruby
+# Reversible: each operation gives Rails what it needs to undo it
+def change
+  remove_column :activities, :note, :string  # type lets Rails re-add the column on rollback
+  drop_table :legacy_widgets do |t|           # block lets Rails recreate the table on rollback
+    t.string :name
+    t.timestamps
+  end
+end
+```
+
+A `down` may be genuinely impractical for a destructive data migration (for
+example, collapsing an array back to a scalar loses information). Document the
+exception in the migration with a comment and define a `down` that restores the
+schema even if it cannot restore every value.
+
+### 3. No model-class references inside migrations
+
+A migration must replay correctly on a fresh deployment install years from now,
+when the app-level model it once referenced may have been renamed or deleted.
+Referencing `Activity` or `ReviewActivityReportTask` inside a migration ties
+that historical migration to today's code; a future rename breaks the replay at
+constant-lookup time.
+
+Use raw SQL, or a migration-local anonymous ActiveRecord class scoped to the table:
+
+```ruby
+# Bad: depends on the Activity constant still existing at replay time
+Activity.where(type: nil).update_all(type: "WorkActivity")
+
+# Good: raw SQL, no app-code dependency
+execute <<~SQL
+  UPDATE activities SET type = 'WorkActivity' WHERE type IS NULL
+SQL
+
+# Good: migration-local class, decoupled from app code
+klass = Class.new(ActiveRecord::Base) { self.table_name = "activities" }
+klass.where(type: nil).update_all(type: "WorkActivity")
+```
+
+### 4. Document required upgrade stops
+
+When a data migration assumes a specific intermediate state (for example, "the
+backfill only works if you are already on `v2026.2.0`"), an implementer who
+jumps several versions at once can land in that state without the migration
+that produced it. Call out the required stop, flagged as a migration note, in
+the release notes for the gating version.
+
+OSCER versions are CalVer (`v2026.N.0`), not SemVer, so there is no "major
+version bump" to signal a breaking upgrade. Required upgrade stops are how a
+breaking migration boundary is communicated instead.
+
+### 5. Deployment-owned migrations and `db/schema.rb`
+
+Deployments add their own migrations for deployment-specific schema (for
+example, adding a `county` column). These are files OSCER itself never ships,
+and Rails migration filenames are timestamp-prefixed, so they don't collide by
+name with OSCER's migrations. A `*_custom_*` marker in the filename (for
+example, `20260301120000_custom_add_county_to_members.rb`) is a *recommended
+naming convention* for keeping deployment migrations visually distinct from
+OSCER's; configure your upstream-sync process to preserve these
+deployment-owned files.
+
+`db/schema.rb` is the one real friction point. Rails regenerates it from the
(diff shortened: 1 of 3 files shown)


docs/reporting-app/migrations.md
```diff
@@ -0,0 +1,221 @@
...
+over, then contract the old shape.
+
+## Fresh setup vs. incremental migration
+
+A fresh database setup and an incremental migration use different database entry
+points, and the distinction matters for replay safety:
+
+| Path | Command | Behavior |
+|------|---------|----------|
+| Fresh setup / boot | `db:prepare` (`bin/docker-entrypoint`, `bin/setup`) | On a fresh DB: creates it and loads `db/schema.rb`, then seeds (never replays the full migration history). On an existing DB: applies only pending migrations (equivalent to `db:migrate`). |
```
jgavin-nava, line 155: Could you mention the `make init-db` command here?

jgavin-nava (approved):
LGTM otherwise

PR #654, opened 2026-06-11
OSCER-638: Allow multiple activity reports

### LINKED ISSUE #638: Allow multiple activity report application forms
## Background
Members need to be able to submit new activity report application forms as long as the verification window has not closed.

## User Story

As a member I want to be able to submit new activity report application forms so that I can exercise my rights within the verification window.

## Acceptance Criteria

- [ ] Member can submit new activity report application forms as long as within verification window
- [ ] Member cannot submit new activity report application forms when case closed
- [ ] Member cannot submit new activity report application forms when verification window closed

## Out-of-scope
Changes to activity report application form wizard are out of scope 

## Open questions 
Does the verification window need to be open to resubmit?

[Tech spec for verification window](https://docs.google.com/document/d/10raEsJU-dEzJ4ExXzcWQXpHWAailVh5UoUm73IIBSAk/edit?tab=t.0)

jdettmannnava:
## Ticket

Resolves #638

## Changes

Allows a member to submit another activity report after a prior one is denied, as long as the certification case is still open and the verification window has not ended. This mirrors the multiple-exemption-forms work (#646 / OSCER-637).

- **`CertificationCase#deny_activity_report`** now keeps the case open while the verification window is open (or has not been opened yet); only a final denial after the window has ended closes the case.
- **Business process**: a non-final `ActivityReportDenied` event routes the member back to `report_activities` instead of `end`.
- **`ActivityReportApplicationForm`**: replaced the in-progress uniqueness validation with create-context `case_not_closed` / `no_pending_forms` validations; added `flow_status` and `has_pending_form`.
- **Dashboard helper**: status checks (`is_activity_report_submitted?/approved?/denied?`) now read the form's `flow_status`; added `should_show_submit_new_activity_report_form?`.
- **Controller**: `create` now matches the exemption controller — a `before_action` builds and saves the form, translating creation errors into friendly notices; `new` guards against a closed case, ended verification window, or an already-pending form.
- **Dashboard denied partial**: added a "Submit new activity report" button (shown only when a resubmission is allowed).
- **Activity report show page**: derives its status message from the form's `flow_status`, so a resubmitted form no longer inherits a prior form's "denied" message.

## Context for reviewers

The key behavioral change is that an activity-report denial is no longer always terminal. While the verification window is open, denial refers the case back to the member (case stays open, member status is `not_compliant`, business process returns to `report_activities`); once the window has ended, denial is final and closes the case.

`flow_status` is the mechanism that keeps per-form status correct once multiple forms exist on a case: it reports the case-level approval status only for the specific form whose review task completed, and `submitted` / `in_progress` otherwise. The dashboard and the show page both use it now, since reading the case-level `activity_report_approval_status` directly would mis-label a freshly resubmitted form as denied.

## Testing

- Full RSpec suite green (2184 examples, 0 failures); `make lint` clean.
- Manually tested:
  - Denied a report (window open) → dashboard shows "Submit new activity report"
  - Submitted a second report → both dashboard and show page show "being reviewed"
  - Originally-denied report still shows its denial
  - "Submit new activity report" button hidden when case closed
  - "Submit new activity report" button hidden when verification window ended
  - Cannot start a new form when case closed → redirected to dashboard with notice
  - Cannot start a new form when verification window ended → redirected to dashboard with notice

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/controllers/activity_report_application_forms_controller.rb
+++ b/reporting-app/app/controllers/activity_report_application_forms_controller.rb
@@ -12,6 +12,7 @@ class ActivityReportApplicationFormsController < ApplicationController
     accept_doc_ai
   ]
   before_action :authenticate_user!
+  before_action :create_activity_report_application_form, only: %i[create]
   before_action :set_certification_case, only: %i[show edit review update]
   before_action :set_certification, only: %i[show review]
   before_action :set_monthly_statistics, only: %i[show review]
@@ -29,15 +30,20 @@ class ActivityReportApplicationFormsController < ApplicationController
       return
     end
     @certification_case = certification_service.find(params[:certification_case_id], hydrate: false)
+
+    if @certification_case.closed?
+      redirect_to dashboard_path, notice: t(".errors.case_closed")
+    elsif @certification_case.verification_window_ended?
+      redirect_to dashboard_path, notice: t(".errors.verification_window_ended")
+    elsif ActivityReportApplicationForm.has_pending_form(@certification_case.id)
+      redirect_to dashboard_path, notice: t(".errors.form_in_progress")
+    end
   end
 
   # POST /activity_report_application_forms
   def create
-    @activity_report_application_form = authorize ActivityReportApplicationForm.new(activity_report_application_form_create_params)
-    @activity_report_application_form.user_id = current_user.id
-
     respond_to do |format|
-      if @activity_report_application_form.save
+      if @activity_report_application_form.valid?
         format.html do
           skip_ai = params.dig(:activity_report_application_form, :skip_ai) == "1"
           session[:doc_ai_skip] = skip_ai
@@ -45,7 +51,6 @@ class ActivityReportApplicationFormsController < ApplicationController
         end
         format.json { render :show, status: :created, location: @activity_report_application_form }
       else
-        flash[:errors] = @activity_report_application_form.errors.full_messages
         format.html { redirect_to dashboard_path }
         format.json { render json: @activity_report_application_form.errors, status: :unprocessable_content }
       end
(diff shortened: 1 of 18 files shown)


reporting-app/app/models/activity_report_application_form.rb
```diff
@@ -60,8 +61,44 @@ def reporting_period_dates
       .map { |ym| Date.new(ym.year, ym.month, 1) }
   end
 
+  def flow_status
```
jgavin-nava, line 64: There is an issue here I believe that not sure if its a problem now but could cause confusion at some point but if a member uploads an activity report and it gets denied and then it uploads a new one and it gets approved the old ones will display the activity report is approved. The statuses should probably be derived per form

reporting-app/app/controllers/activity_report_application_forms_controller.rb
```diff
@@ -29,23 +30,27 @@ def new
...
   end
 
   # POST /activity_report_application_forms
   def create
-    @activity_report_application_form = authorize ActivityReportApplicationForm.new(activity_report_application_form_create_params)
-    @activity_report_application_form.user_id = current_user.id
-
     respond_to do |format|
-      if @activity_report_application_form.save
+      if @activity_report_application_form.valid?
```
jgavin-nava: This validation is probably unnecessary since the `create_activity_report_application_form` already rescues on `save!` so I dont think it can ever be invalid but then again if something weird happens maybe good to have another check

jgavin-nava (commented):
Might be another ticket but now that there is multiple activity reports shown. Here https://github.com/navapbc/oscer/blob/main/reporting-app/app/controllers/certification_cases_controller.rb#L22 loads an abritary form for the staff case show and so does here https://github.com/navapbc/oscer/blob/main/reporting-app/app/controllers/certification_cases_controller.rb#L79

reporting-app/app/models/activity_report_application_form.rb
jdettmannnava: Yup, future problem :)

jdettmannnava:
> Might be another ticket but now that there is multiple activity reports shown. Here https://github.com/navapbc/oscer/blob/main/reporting-app/app/controllers/certification_cases_controller.rb#L22 loads an abritary form for the staff case show and so does here https://github.com/navapbc/oscer/blob/main/reporting-app/app/controllers/certification_cases_controller.rb#L79

Yup, new ticket: https://github.com/navapbc/oscer/issues/656

reporting-app/app/models/activity_report_application_form.rb
jdettmannnava: https://github.com/navapbc/oscer/issues/655

reporting-app/app/controllers/activity_report_application_forms_controller.rb
jdettmannnava: Fixed!

jgavin-nava (approved):

Issue #655, opened 2026-06-12
Make the determination outcome (approved/denied) available per application form


jdettmannnava:
## Background

We need each application form (both activity report and exemption forms) to carry its own
determination outcome — approved or denied — at the data layer, so that the member and
staff UIs can later read the result of each individual form. This ticket is a **backend
change only**; surfacing the outcome in the UI is handled separately.

Today there is no clean path from a form to its own outcome. The decision is recorded in
two places, neither of which is form-specific:

- A `Determination` record whose `subject` is the **`Certification`** (not the form),
  carrying `outcome: compliant | exempt | not_compliant`.
- A **case-level fact** on `CertificationCase` (`activity_report_approval_status` /
  `exemption_request_approval_status`, set to `"approved"`/`"denied"` in
  `CertificationCase#accept_activity_report`, `#deny_activity_report`, and the exemption
  equivalents).

Because that status lives on the **case**, it is not tied to a specific form. This was an
acceptable shortcut while a case could only ever have one form. But we are now adding the
ability to have **multiple forms per case** (one-form-per-case uniqueness was removed in
the `allow_multiple_forms_per_case` migration / OSCER-637), so a single case-level status
can no longer represent the result of each form — it only reflects the **most recent**
decision. A case with a denied Form A and a re-submitted, approved Form B has no way to
surface each form's own outcome. There is no longer an easy mapping from result to form,
which is exactly the gap this story closes.

## User Story

As a developer building the member and staff UIs, I want each application form to expose
its own determination outcome, so that I can display the result of each form on a case —
including earlier forms superseded by a re-submission — without having to infer it from
case-level state.

## Acceptance Criteria

- [ ] Each activity report form exposes its own determination outcome (approved or denied)
      at the data layer.
- [ ] Each exemption form exposes its own determination outcome (approved or denied) at the
      data layer.
- [ ] When a case has multiple forms (e.g. a denied form followed by a re-submission), each
      form reports the outcome of its own decision rather than the case's latest status.
- [ ] A form that has not yet received a determination (draft / in progress / awaiting
      review) reports no outcome, distinguishable from approved/denied.
- [ ] The outcome is populated at the point the decision is made.
- [ ] Forms that already have a determination are backfilled so their outcome is populated
      for historical data.
- [ ] Existing case-level approval-status behavior and any views/helpers that depend on it
      continue to work.

## Out-of-scope (optional)

- Any member- or staff-facing UI changes to display the outcome (separate ticket).
[...]

PR #658, opened 2026-06-12
Jdettmann/oscer 655 approval status migration

jdettmannnava:
## Ticket

Part of #655 (schema prerequisite — does not resolve the ticket on its own)

## Changes

- Add a nullable `approval_status` column to `strata_tasks` so review tasks can record their staff decision (`approved` / `denied`; `nil` = undecided).
- Add an index on `strata_tasks.application_form_id` (the column was added in OSCER-585 without one).
- Backfill `approval_status` for existing review tasks from the case-level approval facts.

## Context for reviewers

This is a schema-only PR, split out from the feature work. Tasks were chosen to hold the value, as application forms are immutable after submission.

**Backfill rule.** A case can now hold multiple forms (and review tasks) after a denial + re-submission, but the case-level fact (`activity_report_approval_status` / `exemption_request_approval_status`) only retains the most recent decision. So per case, the most-recent review task inherits the case fact and every superseded task is set to `denied` (a later form only exists because the earlier one was denied). The backfill is idempotent and logs any case with more than one task.

## Testing

- `make db-migrate` applied cleanly; backfill logged the multi-task cases it touched.
- Spot-checked the backfilled data for both form types: in multi-task cases only the most-recent task carries the `approved` outcome, with the superseded tasks set to `denied` — confirming the attribution rule.
- `make lint` — no offenses.
- `make test` — 2184 examples, 0 failures (96.05% line / 82.04% branch coverage).

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/db/migrate/20260612120000_add_approval_status_to_strata_tasks.rb
+++ b/reporting-app/db/migrate/20260612120000_add_approval_status_to_strata_tasks.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+class AddApprovalStatusToStrataTasks < ActiveRecord::Migration[8.0]
+  def up
+    add_column :strata_tasks, :approval_status, :string
+    # application_form_id was added (OSCER-585) without an index; the per-form outcome
+    # delegation now looks tasks up by it, so index it here.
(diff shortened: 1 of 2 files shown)


giverm (approved):
LGTM

PR #659, opened 2026-06-15
OSCER-655 Make decision available via form

### LINKED ISSUE #655: Make the determination outcome (approved/denied) available per application form

jdettmannnava:
## Ticket

Resolves #655

## Changes

Make each application form's own determination outcome (approved / denied) available at the data layer, so the member and staff UIs can later read the result of each individual form.

- **Review tasks** record the staff decision via an `approval_status` enum (`approved` / `denied`; `nil` until decided) on `ReviewActivityReportTask` and `ReviewExemptionClaimTask`.
- **Application forms** expose their own outcome by delegating to their bound review task: a new `FormApprovalStatus` concern adds `approval_status` / `approved?` / `denied?` plus a `has_review_task` macro that wires `has_one :review_task` on each form.
- **Review-task controllers** set `@task.approval_status` when staff approve or deny; the existing `@task.completed!` persists it.

## Context for reviewers

The form already had a determination concept, but nothing tied an outcome to a *specific* form — the decision lived only on the `Certification`'s `Determination` and on a case-level approval fact, neither of which is form-specific. With multiple forms per case now possible (OSCER-637 / OSCER-638), a denied form followed by an approved re-submission could not report each form's own result.

Key design decisions:
- **Outcome lives on the review task, not the form.** `Strata::ApplicationForm` is immutable once submitted (`prevent_changes_if_submitted`), and the decision is really a property of the review, not of the member's submitted data. Storing it on the task (which OSCER-585 already binds to its form) keeps the form immutable; the form delegates. This is also more correct for multi-form cases — each form points at its own task's decision rather than the shared, last-write-wins case fact.
- **Case-level approval facts are untouched.** All existing views/helpers that read `activity_report_approval_status` / `exemption_request_approval_status` continue to work; this PR only adds the per-form path.
- **`nil` is a distinct state** — a form whose task is undecided (or has no task) reports no outcome, distinguishable from approved/denied.

Out of scope: surfacing the outcome in the member/staff UI (separate ticket). Note OSCER-638 added an activity-report `flow_status` that derives a similar status from the completed task + case fact; the two coexist, and consolidating them into a single source of truth is a reasonable follow-up.

## Testing

- New specs (red → green) cover the task enum, the form delegation (no task / undecided / approved / denied), and the controller flow stamping both the task and — through delegation — the form.
- `make lint` — no offenses.
- `make test` — 2202 examples, 0 failures (96.07% line / 82.07% branch coverage).

**Manual testing.** Ran the staff review flow locally and read each form's outcome on its `show` page, confirming `form.approval_status` is persisted and reflects the form's own decision:

- Exemption — **deny** → form reads `denied`; **approve** → form reads `approved`.
- Activity report — **deny** → form reads `denied`; **approve** → form reads `approved`.
- **Re-submission (multi-form):** after denying form A and approving its replacement form B, form A still reads `denied` while form B reads `approved` — each form reports its own decision, where the case-level fact only reflects the latest.
- **Request more information (no decision):** form keeps no outcome (`nil`) — only an approve/deny records one.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/concerns/form_approval_status.rb
+++ b/reporting-app/app/models/concerns/form_approval_status.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+# Exposes an application form's own determination outcome (approved/denied) without
+# mutating the immutable form. The decision is recorded on the form's review task; the
+# form simply delegates to it. A form with no review task, or whose review task is still
+# undecided, reports no outcome (nil), distinguishable from approved/denied.
+#
+# Including models must declare a +review_task+ association (the bound
+# ReviewActivityReportTask / ReviewExemptionClaimTask).
+module FormApprovalStatus
+  extend ActiveSupport::Concern
+
+  def approval_status = review_task&.approval_status
+  def approved? = approval_status == "approved"
+  def denied? = approval_status == "denied"
+end
--- a/reporting-app/app/models/concerns/form_approval_status.rb
+++ b/reporting-app/app/models/concerns/form_approval_status.rb
@@ -22,6 +22,6 @@ module FormApprovalStatus
   end
 
   def approval_status = review_task&.approval_status
-  def approved? = approval_status == "approved"
-  def denied? = approval_status == "denied"
+  def approved? = review_task&.approved?
+  def denied? = review_task&.denied?
 end
--- a/reporting-app/app/models/concerns/form_approval_status.rb
+++ b/reporting-app/app/models/concerns/form_approval_status.rb
@@ -5,10 +5,21 @@
 # form simply delegates to it. A form with no review task, or whose review task is still
 # undecided, reports no outcome (nil), distinguishable from approved/denied.
 #
-# Including models must declare a +review_task+ association (the bound
-# ReviewActivityReportTask / ReviewExemptionClaimTask).
+# Including models bind their review task with +has_review_task+, supplying the bound
+# ReviewActivityReportTask / ReviewExemptionClaimTask class.
 module FormApprovalStatus
   extend ActiveSupport::Concern
+
+  class_methods do
+    # Each form binds to its own review-task subclass; the association config is shared.
+    def has_review_task(class_name)
+      has_one :review_task,
+        class_name: class_name,
+        foreign_key: :application_form_id,
+        inverse_of: :application_form,
+        strict_loading: false
+    end
+  end
 
   def approval_status = review_task&.approval_status
   def approved? = approval_status == "approved"
--- a/reporting-app/Gemfile.lock
+++ b/reporting-app/Gemfile.lock
@@ -341,7 +341,7 @@ GEM
     nenv (0.3.0)
     net-http (0.9.1)
       uri (>= 0.11.1)
-    net-imap (0.6.4)
+    net-imap (0.6.4.1)
       date
       net-protocol
     net-pop (0.1.2)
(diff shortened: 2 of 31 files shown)


reporting-app/app/models/concerns/form_approval_status.rb
```diff
@@ -0,0 +1,27 @@
...
+        class_name: class_name,
+        foreign_key: :application_form_id,
+        inverse_of: :application_form,
+        strict_loading: false
+    end
+  end
+
+  def approval_status = review_task&.approval_status
+  def approved? = approval_status == "approved"
+  def denied? = approval_status == "denied"
```
giverm: Since you already have the `enum` defined on the review task models, we could do something like


```suggestion
  def approved? = !!review_task&.approved?
  def denied? = !!review_task&.denied?
```

giverm (approved):
Looks great!

PR #664, opened 2026-06-15
OSCER-656 show all submitted activity forms in case view

### LINKED ISSUE #656: Update staff certification case show to handle multiple forms of each type
## Background

The staff certification case detail page (`CertificationCasesController#show`) was built
when a case could only have one activity report form and one exemption form. We are now
allowing **multiple forms of each type per case** — exemptions in OSCER-637, and activity
reports in OSCER-638 (`jdettmann/oscer-638-multiple-activity-reports`). That work updated
the member-facing flow but did **not** touch the staff case show, so the staff view still
assumes a single form of each type and will silently surface only one.

Concretely, the single-form assumptions in the staff show path are:

- `app/controllers/certification_cases_controller.rb:22` loads a singular
  `@activity_report = ActivityReportApplicationForm.find_by(certification_case_id: @case.id)`.
- The `application_form_ids` helper (`certification_cases_controller.rb:77–81`) calls
  `find_by_certification_case_id` for each form type, returning at most one form per type,
  so `@information_requests` only reflects a single activity report and a single exemption
  form.
- The view branches on the singular `@activity_report`
  (`app/views/certification_cases/show.html.erb` and
  `app/views/certification_cases/_external_data_section.html.erb`).

When a case has more than one form of a type (e.g. a denied activity report followed by a
re-submission), staff cannot see all of them — only one is loaded and rendered, and
information requests tied to the other forms are dropped.

## User Story

As a staff member reviewing a certification case, I want to see all of the activity report
and exemption forms on the case, so that I can review the full submission history when a
case has more than one form of a type.

## Acceptance Criteria

- [ ] The case show action loads all activity report forms and all exemption forms for the
      case, not just one of each type.
- [ ] The case show view displays each form on the case, including multiple forms of the
      same type.
- [ ] Information requests are loaded and displayed for every form on the case, not just
      one form per type.
- [ ] A case with exactly one form of a type renders the same as it does today (no
      regression for the common case).
- [ ] A case with no forms of a type still renders its appropriate empty state.
- [ ] Each displayed form is distinguishable (e.g. by submission date / status) so staff
      can tell re-submissions apart.

## Out-of-scope (optional)

- Member-facing multiple-forms behavior (OSCER-637, OSCER-638).
- Exposing a per-form determination outcome at the data layer (separate backend ticket).
- Changes to how determinations or approval status are computed.

## Dependencies & Risks (optional)

- Depends on / coordinates with OSCER-638
  (`jdettmann/oscer-638-multiple-activity-reports`), which introduces
  `ActivityReportApplicationForm#flow_status` and `.has_pending_form` and removes the
  in-progress uniqueness constraint.
- Builds on OSCER-637 (multiple exemption forms).
- The `find_by_certification_case_id` finders and `application_form_ids` helper are
  single-form by design; other call sites may rely on them and should be checked before
  changing shared methods.
- Determining how to present multiple same-type forms in the existing layout may need
  design input.

## Open questions (optional)

- How should multiple same-type forms be ordered and presented in the staff view (most
  recent first? grouped by status?) — does this need design?
- Should superseded/denied forms be visually de-emphasized vs. the active form?
- Per-form approval status: the member side now reads `flow_status` per form; should the
  staff view use the same per-form status rather than the case-level
  `activity_report_approval_status` / `exemption_request_approval_status` facts?

- `app/controllers/certification_cases_controller.rb` (`#show`, `application_form_ids`)
- `app/views/certification_cases/show.html.erb`
- `app/views/certification_cases/_external_data_section.html.erb`
- `app/models/activity_report_application_form.rb`, `app/models/exemption_application_form.rb`
- Branch: `jdettmann/oscer-638-multiple-activity-reports` (OSCER-638)
- `tmp/multiple-application-forms-per-case.md`

jdettmannnava:
## Ticket

Resolves #656

## Changes

Update the staff certification case detail page (`CertificationCasesController#show`) to handle
**multiple forms of each type per case**. The page was built when a case could only have one
activity report and one exemption form, so it loaded a single `@activity_report` and surfaced at
most one form (and one set of information requests) per type.

- **Controller loads every activity report form** (`@activity_reports`, most-recent first). The
  most-recent form becomes `@activity_report`, which drives the existing case-level compliance
  summary and the External Data section unchanged.
- **Information requests load for all forms** of both types — `application_form_ids` now returns
  every form's id (`where(...).pluck(:id)`) instead of one per type, so requests tied to an older
  activity report or to an exemption form are no longer dropped.
- **DocAI confidence spans all forms** — confidence is now built from the activities of every
  activity report on the case, not just one.
- **View renders line items per form**, each labeled with a status tag + submission date when
  there is more than one form, so re-submissions are distinguishable. The status tag reads each
  form's **own** decision via the per-form `approval_status` (#659), falling back to the form's
  workflow status when undecided.

## Context for reviewers

OSCER-637 (multiple exemption forms) and OSCER-638 (multiple activity reports) removed the
one-form-per-case constraint on the member side but left the staff view single-form. A case with
a denied activity report followed by a re-submission would show staff only one form, and
information requests tied to the others were silently dropped.

Key design decisions:
- **Single-form behavior is preserved exactly.** `@activity_report` is now the most-recent form
  (where there's only one, it's that one), so the compliance summary and the
  `_external_data_section` "summarized under Activity Report" branching render byte-for-byte as
  before. The per-form line-items loop only adds a status/date header when `size > 1`; a lone
  form still shows line items only under DocAI, with no header.
- **Per-form status, not the case-level fact.** Each form's status tag derives from its own
  review-task decision (`approval_status`, #659) rather than the shared, last-write-wins
  case-level `activity_report_approval_status` that `flow_status` reads — so on a multi-form case
  a denied form and its approved re-submission each show their own outcome.
- **Compliance computation is untouched.** The hours/income summary remains a case-level roll-up
  driven by the representative (most-recent) form — this ticket changes display only, not how
  determinations or totals are computed.
- **Exemption form display is out of scope.** Exemptions aren't rendered on the staff show page
  today; this PR only ensures their information requests load.

## Testing

- New request specs (red → green) cover multiple activity report forms (both forms' line items
  render, distinguishable by status, works with DocAI off, compliance summary still renders) and
  information requests spanning more than one form. Existing single-form examples are untouched.
- `make lint` — no offenses.
- `make test` — 2207 examples, 0 failures (96.03% line / 82.1% branch coverage).

**Manual testing.** Verified locally on the staff case show page:
- **Single activity report** — renders as before (no regression).
- **Multiple activity reports** — both forms' line items render, distinguishable by status/date.
- **Multiple information requests** — requests for every form on the case are listed.

many_forms

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/views/certification_cases/show.html.erb
+++ b/reporting-app/app/views/certification_cases/show.html.erb
@@ -127,9 +127,10 @@
       <h3 class="margin-top-4"><%= t(".activity_line_items") %></h3>
       <% @activity_reports.each do |form| %>
         <% if @activity_reports.size > 1 %>
+          <% form_status = activity_report_display_status(form) %>
           <h4 class="margin-top-3">
             <%= t(".form_submitted_on", date: (form.submitted_at || form.updated_at).strftime("%m/%d/%Y")) %>
-            <span class="usa-tag <%= activity_report_status_class(form.flow_status) %>"><%= t(".form_status.#{form.flow_status}") %></span>
+            <span class="usa-tag <%= activity_report_status_class(form_status) %>"><%= t(".form_status.#{form_status}") %></span>
           </h4>
         <% end %>
         <%= render partial: "activity_report_application_forms/staff_activity_report", locals: { activity_report: form } %>
--- a/reporting-app/app/views/certification_cases/show.html.erb
+++ b/reporting-app/app/views/certification_cases/show.html.erb
@@ -122,9 +122,18 @@
     <% else %>
       <p><%= t(".no_activity_report") %></p>
     <% end %>
-    <% if @activity_report && Features.doc_ai_enabled? %>
+    <% show_line_items = @activity_reports.size > 1 || Features.doc_ai_enabled? %>
+    <% if show_line_items && @activity_reports.any? %>
       <h3 class="margin-top-4"><%= t(".activity_line_items") %></h3>
-      <%= render partial: "activity_report_application_forms/staff_activity_report", locals: { activity_report: @activity_report } %>
+      <% @activity_reports.each do |form| %>
+        <% if @activity_reports.size > 1 %>
+          <h4 class="margin-top-3">
+            <%= t(".form_submitted_on", date: (form.submitted_at || form.updated_at).strftime("%m/%d/%Y")) %>
+            <span class="usa-tag <%= activity_report_status_class(form.flow_status) %>"><%= t(".form_status.#{form.flow_status}") %></span>
+          </h4>
+        <% end %>
+        <%= render partial: "activity_report_application_forms/staff_activity_report", locals: { activity_report: form } %>
+      <% end %>
     <% end %>
   <% end %>
 <% end %>
--- a/reporting-app/app/views/certification_cases/show.html.erb
+++ b/reporting-app/app/views/certification_cases/show.html.erb
@@ -122,14 +122,14 @@
     <% else %>
       <p><%= t(".no_activity_report") %></p>
     <% end %>
-    <% show_line_items = @activity_reports.size > 1 || Features.doc_ai_enabled? %>
-    <% if show_line_items && @activity_reports.any? %>
+    <% show_line_items = @activity_reports.size > 1 || (@activity_reports.any? && Features.doc_ai_enabled?) %>
+    <% if show_line_items %>
       <h3 class="margin-top-4"><%= t(".activity_line_items") %></h3>
       <% @activity_reports.each do |form| %>
         <% if @activity_reports.size > 1 %>
           <% form_status = activity_report_display_status(form) %>
           <h4 class="margin-top-3">
-            <%= t(".form_submitted_on", date: (form.submitted_at || form.updated_at).strftime("%m/%d/%Y")) %>
+            <%= t(".form_submitted_on", date: l(form.submitted_at || form.updated_at)) %>
             <span class="usa-tag <%= activity_report_status_class(form_status) %>"><%= t(".form_status.#{form_status}") %></span>
           </h4>
         <% end %>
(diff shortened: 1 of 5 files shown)


reporting-app/app/views/certification_cases/show.html.erb
```diff
@@ -122,9 +122,19 @@
...
-    <% if @activity_report && Features.doc_ai_enabled? %>
+    <% show_line_items = @activity_reports.size > 1 || (@activity_reports.any? && Features.doc_ai_enabled?) %>
+    <% if show_line_items %>
       <h3 class="margin-top-4"><%= t(".activity_line_items") %></h3>
-      <%= render partial: "activity_report_application_forms/staff_activity_report", locals: { activity_report: @activity_report } %>
+      <% @activity_reports.each do |form| %>
+        <% if @activity_reports.size > 1 %>
+          <% form_status = activity_report_display_status(form) %>
+          <h4 class="margin-top-3">
+            <%= t(".form_submitted_on", date: l(form.submitted_at || form.updated_at)) %>
```
giverm, line 132: This will always display "Submitted {date}", including when the application form is still `in_progress`. Probably not a huge deal, since the status is reported clearly next to the date, but I'd consider conditionally render something like "Started {date}" when the form is `in_progress`

giverm (approved):
Really nice work! One small UI comment, but otherwise this looks great.

reporting-app/app/views/certification_cases/show.html.erb
jdettmannnava: Hrmm. I actually like 'submitted' here, as it is the "update" date, which means the last time it was worked on. "Started" would be created_at, which strikes me as less useful.

giverm: That works for me. It just struck me as odd at first glance to see a submission date for a form that has not reached the `submitted` state yet. But the status display makes it clear, so :shipit:!

PR #665, opened 2026-06-15
Add pagination to the staff certification cases index

### LINKED ISSUE #663: Add pagination to the staff certification cases index
## Background

The staff certification cases index (`CertificationCasesController#index` / `#closed`, rendered via the Strata `IndexComponent`) loads **every** matching case for the staff user's region in one query, then hydrates each with its `Certification`. On the AWS dev environment this page has become slow to load as case volume grows.

The query is currently unbounded and, for the **open** tab, **unordered**: the `open` scope is the enum-generated `where(status: :open)` with no `ORDER BY`. The apparent "oldest-first" ordering staff see is PostgreSQL heap return order, not application logic, so it is not guaranteed. (The `closed` scope, by contrast, is explicitly `order(created_at: :desc)` in the SDK.)

## Goal

Paginate both the open and closed tabs so each request loads and hydrates only one page of cases, and give the open tab a deterministic sort order.

## Design decisions

- **Library: Pagy.** Lightweight, no ActiveRecord monkeypatching (keeps `CertificationCase` clean, consistent with the codebase's deliberate aggregate boundaries), minimal runtime overhead. Kaminari's view-helper advantage is moot here because pagination controls must be hand-written as USWDS markup regardless.
- **Sort order: newest first** (`order(created_at: :desc, id: :desc)`). The `id` tiebreaker makes the order total so `OFFSET` paging is stable. This also aligns the open tab with the closed tab's existing intentional `created_at DESC` order.
- **Pagination applied before `CertificationService#hydrate_cases_with_certifications!`** so only the current page's certifications are loaded.
- **Controls live in a reusable USWDS partial** (`app/views/shared/_pagination.html.erb`) driven by the Pagy object, rendered around the untouched Strata `IndexComponent`. All strings go through i18n (en + es-US).

## Acceptance criteria

- [ ] Open and closed staff cases index tabs are paginated (default 25 per page).
- [ ] Open tab is ordered newest-first (`created_at DESC, id DESC`); paging is stable across requests (no repeated/skipped rows).
- [ ] Only one page of cases is loaded and hydrated per request (no full-table load).
- [ ] Pagination controls use USWDS markup and are internationalized (en + es-US).
- [ ] Out-of-range `?page=` values are handled gracefully (clamp to last page).
- [ ] Request specs cover: page count, per-page limit, ordering, and an out-of-range page param.

## Follow-up / note

For large regions the `COUNT` and page queries still hit the unindexed `status` column behind the region join. Consider an index supporting `(status, created_at)` (scoped appropriately) as a fast-follow; track separately if it grows beyond this story.

giverm:
## Ticket

Resolves #663

## Summary

The staff certification cases index loaded every case for the signed-in user's region in one unbounded query and hydrated each with its Certification, so the page got slower as case volume grew on the dev environment. This paginates both the open and closed tabs (25 per page) and gives the open tab a deterministic sort order, so each request loads and hydrates only one page of cases.

## Changes

- Add the `pagy` gem and a `config/initializers/pagy.rb` setting a default of 25 per page and clamping out-of-range `?page=` values to the last page.
- Paginate `CertificationCasesController#index` and `#closed`, applying the page limit before `hydrate_cases_with_certifications!` so only the current page's certifications are loaded.
- Order the open tab newest-first (`created_at DESC`) with an `id` tiebreaker; add the same tiebreaker to the closed tab (which the SDK already orders `created_at DESC`).
- Add a reusable `shared/_pagination` partial that renders USWDS pagination markup from the Pagy object, rendered around the unchanged Strata `IndexComponent`.
- Add `include Pagy::Backend` (controllers) and `include Pagy::Frontend` (helpers).
- Add en locale keys for the pagination controls.

## Notes

- The open tab previously had no `ORDER BY`; the apparent oldest-first ordering was PostgreSQL heap return order, not application logic, so it was unstable and unsafe for OFFSET paging. Newest-first also aligns it with the closed tab's existing intentional order.
- Pagy was chosen over Kaminari for its smaller footprint and because it does not monkeypatch ActiveRecord. The view-helper convenience Kaminari offers does not apply here, since the controls are hand-written USWDS markup either way.
- For large regions the `COUNT` and page queries still scan the unindexed `status` column behind the region join. A `(status, created_at)` index is a reasonable fast-follow and is tracked separately.

## Testing

- `make test`: 2209 examples, 0 failures (96.07% line / 82.12% branch coverage). New request specs cover the pagination behavior.
- `make lint`: no offenses.
- Manually verified the rendered page: with 31 open cases in one region, the USWDS pager shows two pages, page 1 lists the 25 newest with no Previous link, page 2 lists the remainder with no Next link, and the control is absent when results fit on one page. 
Screenshot 2026-06-15 at 3 41 17 PM

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/controllers/certification_cases_controller.rb
+++ b/reporting-app/app/controllers/certification_cases_controller.rb
@@ -7,12 +7,14 @@ class CertificationCasesController < StaffController
   before_action :set_certification, only: %i[ show tasks documents notes ]
 
   def index
-    @cases = policy_scope(CertificationCase).open
+    # Newest first, with an id tiebreaker so the order is total and OFFSET paging is stable.
+    @pagy, @cases = pagy(policy_scope(CertificationCase).open.order(created_at: :desc, id: :desc))
     certification_service.hydrate_cases_with_certifications!(@cases)
   end
 
   def closed
-    @cases = policy_scope(CertificationCase).closed
+    # The closed scope already orders by created_at :desc; add the id tiebreaker for stable paging.
+    @pagy, @cases = pagy(policy_scope(CertificationCase).closed.order(id: :desc))
     certification_service.hydrate_cases_with_certifications!(@cases)
     render :index
   end
--- a/reporting-app/app/controllers/certification_cases_controller.rb
+++ b/reporting-app/app/controllers/certification_cases_controller.rb
@@ -13,7 +13,8 @@ class CertificationCasesController < StaffController
   end
 
   def closed
-    # The closed scope already orders by created_at :desc; add the id tiebreaker for stable paging.
+    # `closed` is defined in the Strata SDK (Strata::Case) and already orders by
+    # created_at :desc; we only add the id tiebreaker here for stable OFFSET paging.
     @pagy, @cases = pagy(policy_scope(CertificationCase).closed.order(id: :desc))
     certification_service.hydrate_cases_with_certifications!(@cases)
     render :index
--- a/reporting-app/app/views/shared/_pagination.html.erb
+++ b/reporting-app/app/views/shared/_pagination.html.erb
@@ -15,6 +15,10 @@
         </li>
       <% end %>
 
+      <%# pagy.series(size: 7) returns one entry per slot, signaled by type:
+          Integer -> a page to link to; String -> the current page; :gap -> an elision ("…").
+          size >= 7 is what enables the first/last anchoring and :gap markers; 7 is also
+          Pagy's default, passed explicitly here to keep that behavior legible at the call site. %>
       <% pagy.series(size: 7).each do |item| %>
         <% if item == :gap %>
           <li class="usa-pagination__item usa-pagination__overflow" role="presentation">
--- a/reporting-app/app/views/shared/_pagination.html.erb
+++ b/reporting-app/app/views/shared/_pagination.html.erb
@@ -1,0 +1,45 @@
+<%# Reusable USWDS pagination controls driven by a Pagy object.
+    Renders nothing unless there is more than one page.
+    See https://designsystem.digital.gov/components/pagination/ %>
+<% if pagy && pagy.pages > 1 %>
+  <nav aria-label="<%= t("pagination.aria_label") %>" class="usa-pagination">
+    <ul class="usa-pagination__list">
+      <% if pagy.prev %>
+        <li class="usa-pagination__item usa-pagination__arrow">
+          <%= link_to pagy_url_for(pagy, pagy.prev),
+                class: "usa-pagination__link usa-pagination__previous-page",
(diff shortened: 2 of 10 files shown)


reporting-app/app/controllers/certification_cases_controller.rb
```diff
@@ -7,12 +7,14 @@ class CertificationCasesController < StaffController
...
-    @cases = policy_scope(CertificationCase).open
+    # Newest first, with an id tiebreaker so the order is total and OFFSET paging is stable.
+    @pagy, @cases = pagy(policy_scope(CertificationCase).open.order(created_at: :desc, id: :desc))
     certification_service.hydrate_cases_with_certifications!(@cases)
   end
 
   def closed
-    @cases = policy_scope(CertificationCase).closed
+    # The closed scope already orders by created_at :desc; add the id tiebreaker for stable paging.
+    @pagy, @cases = pagy(policy_scope(CertificationCase).closed.order(id: :desc))
```
jgavin-nava, line 18: Should this also use order by created_at as well?

reporting-app/app/views/shared/_pagination.html.erb
```diff
@@ -0,0 +1,45 @@
...
+          <%= link_to pagy_url_for(pagy, pagy.prev),
+                class: "usa-pagination__link usa-pagination__previous-page",
+                aria: { label: t("pagination.previous") } do %>
+            <%= uswds_icon("navigate_before") %>
+            <span class="usa-pagination__link-text"><%= t("pagination.previous") %></span>
+          <% end %>
+        </li>
+      <% end %>
+
+      <% pagy.series(size: 7).each do |item| %>
```
jdettmannnava, line 22: Maybe a comment on the conventions of :gap | int | string so not everyone has to go down the rabbit hole? (Like I just enjoyed doing...)

jdettmannnava (approved):
Looks great!

jgavin-nava (approved):
Other than that one question it looks good

reporting-app/app/controllers/certification_cases_controller.rb
giverm: The `closed` scope, which is defined in the SDK, already orders by `created_at: :desc`. I briefly considered adding an `open` scope to do the same, but I think the better move would be to remove the ordering and let consumers of the SDK do the ordering instead.

Anyways, I updated the comment to make it more clear and am going to propose the change in the SDK. If that lands, I'll update the ordering and comment here.

reporting-app/app/views/shared/_pagination.html.erb
giverm: Sorry for the confusion! Comment added

PR #676, opened 2026-06-16
Upgrade pagy to v43 and add a USWDS pagination style

giverm:
## Ticket

No linked issue. Supersedes the Dependabot bump in #674.

## Changes

- Bump `pagy` `9.4.0` to `43.5.6` (`Gemfile`, `Gemfile.lock`).
- Migrate the initializer to the v43 config API: `Pagy::DEFAULT` becomes `Pagy::OPTIONS` (frozen), set `:limit` and `:slots`, and drop the removed `require "pagy/extras/overflow"`.
- Swap `include Pagy::Backend` for `include Pagy::Method` in `ApplicationController` (the controller `pagy` method moved).
- Remove the now-removed `include Pagy::Frontend` from `ApplicationHelper`.
- Add a custom `:uswds` Pagy style (`uswds_series_nav` plus `uswds_page_item` / `uswds_arrow_item` helpers) in `config/initializers/pagy.rb`.
- Replace the v9 `_pagination` partial internals with a single `pagy.series_nav(:uswds, ...)` call, passing the prev/next chevron icons in from the view.
- Update the out-of-range pagination request spec to assert v43's empty-page-with-back-link behavior.

## Context for reviewers

Pagy 43 is a ground-up redesign, not a routine major: the version leap from 9 signals a full API rewrite, so the small Dependabot lockfile diff in #674 hides a breaking change that fails to boot (`LoadError` on the removed `pagy/extras/overflow`). This upgrades pagy and migrates the integration in one atomic PR, since there is no forward-compatible middle state.

The redesign removed three things the app relied on, each fixed here: the `extras/overflow` require, the `Pagy::Backend` controller mixin (now `Pagy::Method`), and the `Pagy::Frontend` view mixin (helpers are now instance methods on `@pagy`). It also dropped the `:last_page` overflow clamp, so an out-of-range page like `?page=9999` now renders an empty page whose pager links back to the last real page via `@pagy.previous`, rather than clamping to it. We accepted that as the new behavior (it still avoids the 500 the original clamp was guarding against).

Pagy ships Bootstrap/Bulma/Tailwind series styles but not USWDS, so rather than monkeypatch or reach into pagy's now-protected `series`, the pager uses pagy's blessed extension point: a custom `:uswds` style. It runs in the Pagy instance context (no ActionView), so it builds markup from `#page_url`/`#series`, reads copy from the fully-qualified `::I18n` (a bare `I18n` inside `class Pagy` resolves lexically to `Pagy::I18n`), and receives the chevron `<svg>` icons from the view, where `uswds_icon`'s `asset_path` is available.

To exercise locally: seed more than 25 cases in your staff region and visit the staff certification cases index (and `/closed`); a middle page shows the gap (ellipsis) markers, and `?page=9999` shows the empty-page behavior described above.

## Testing

- `bundle exec rspec spec/requests/`: 451 examples, 0 failures (full request suite; the app-wide controller/helper mixin changes are exercised broadly).
- `bundle exec rubocop`: clean on all changed Ruby files.
- Browser-verified the rendered pager: USWDS markup, member names, prev/next chevron icons, and gap (ellipsis) markers across a 10-page open list; the closed tab pager; and the out-of-range empty-page behavior.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/Gemfile
+++ b/reporting-app/Gemfile
@@ -21,7 +21,7 @@ gem "sprockets-rails"
 gem "pg", "~> 1.6"
 
 # Pagination for index pages [https://github.com/ddnexus/pagy]
-gem "pagy", "~> 9.3"
+gem "pagy", "~> 43.5"
 
 # Use the Puma web server [https://github.com/puma/puma]
 gem "puma", ">= 5.0"
--- a/reporting-app/Gemfile.lock
+++ b/reporting-app/Gemfile.lock
@@ -411,7 +411,10 @@ GEM
       webfinger (~> 2.0)
     orm_adapter (0.5.0)
     ostruct (0.6.3)
-    pagy (9.4.0)
+    pagy (43.5.6)
+      json
+      uri
+      yaml
     parallel (2.1.0)
     parser (3.3.11.1)
       ast (~> 2.4.1)
@@ -676,6 +679,7 @@ GEM
     websocket-extensions (0.1.5)
     xpath (3.2.0)
       nokogiri (~> 1.8)
+    yaml (0.4.0)
     yard (0.9.43)
     zeitwerk (2.8.2)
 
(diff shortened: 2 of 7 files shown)


jdettmannnava (approved):
LGTM!

Issue #657, opened 2026-06-12
Denial response: backend domain layer (no UI)


jdettmannnava:
## Background
When a member's certification case is denied, their only current way to act during the
verification window is to submit an entirely new activity report. We want a lighter-weight
option: a "denial response" in which the member provides a short written explanation and
optional supporting documents to resolve the denial, which a staff reviewer then approves or
denies.

This story builds only the backend/domain layer for that flow — the data it stores, how
submitting it moves the case into review, and what approving or denying it does to the case —
with no member- or staff-facing screens. Landing the domain layer first lets us validate the
state changes and case outcomes in isolation; the member submission UI and the staff review UI
follow in later stories. It mirrors the existing activity-report and exemption flows and builds
on the verification-window-aware denial groundwork already in place.

## User Story
As the team building the denial-response feature, we want the denial-response domain model,
its review task, and the case approve/deny behavior in place and fully testable, so that the
member and staff interfaces can be added later as thin layers over verified behavior.

## Acceptance Criteria
- [ ] A denial response can be created for a certification case and holds a member comment and
      optional supporting documents, following the standard application-form lifecycle
      (in progress → submitted).
- [ ] A denial response cannot be created when the case is closed or its verification window has
      ended, and cannot be created when a denial-response review is already pending for the case.
- [ ] Submitting a denial response moves the case into a denial-response review state and creates
      a staff review task tied to the submitted response.
- [ ] Approving a denial response resolves the case as compliant, records a determination, and
      closes the case.
- [ ] Denying a denial response keeps the case open while the verification window is open (the
      member may submit again) and closes the case once the window has ended; the outcome event
      that is published reflects which of the two occurred.
- [ ] The denial response exposes a derived status reflecting its latest review outcome,
      consistent with the activity-report and exemption forms.
- [ ] No member- or staff-facing UI is added in this story; behavior is verified via automated
      tests and the Rails console.

## Out-of-scope
- Member submission screens, routes, authorization, and the denied-dashboard entry point /
  "one pending review at a time" gating — later story.
- Staff review/decision screens — later story.
- The "request more information" flow on a denial response — later story.
- Whether an exemption denial should open a verification window — separate decision story. (With
  no window on the exemption side today, an exemption-originated denial response's deny path stays
  in the "window open" branch; acceptable for this backend story.)

## Dependencies & Risks
- Builds on the verification-window-aware activity-report denial groundwork
[...]

PR #677, opened 2026-06-17
OSCER-657 denial response: migration

jdettmannnava:
## Ticket

Step one of #657

## Changes

Add the `denial_response_application_forms` table — the storage layer for the new denial-response
flow, in which a member resolves a denied certification case with a short written comment plus
optional supporting documents.

- New migration `CreateDenialResponseApplicationForms`
- `db/schema.rb` regenerated (schema version bumped to `2026_06_16_170507`).

## Context for reviewers

This is the **migration-only** PR at the base of a three-PR stack for the denial-response backend.
It is split out on its own per the e2e constraint that migrations and the code that uses them ship
in separate PRs. The table columns intentionally mirror the existing `exemption_application_forms`
and `activity_report_application_forms` shapes, since `DenialResponseApplicationForm` (next PR in
the stack) is a `Strata::ApplicationForm` like those.

No model, validations, or behavior land here — those arrive in the follow-up
`denial-response-form-and-process` PR, where they can be tested against this table.

If anyone has a better name than `DenialResponseApplicationForm`, please let me know!

## Testing

- Schema-only change. `make db-migrate` applies cleanly and `make db-reset` rebuilds from
  `schema.rb` without drift.
- `make lint` — no offenses.
- `make test` — 2248 examples, 0 failures (96.03% line / 82.37% branch coverage).

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/db/migrate/20260616170507_create_denial_response_application_forms.rb
+++ b/reporting-app/db/migrate/20260616170507_create_denial_response_application_forms.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class CreateDenialResponseApplicationForms < ActiveRecord::Migration[8.0]
+  def change
+    create_table :denial_response_application_forms, id: :uuid do |t|
+      t.uuid :user_id
+      t.integer :status
+      t.datetime :submitted_at
+      t.uuid :certification_case_id
+      t.text :comment
+
+      t.timestamps
+    end
+
+    add_index :denial_response_application_forms, :certification_case_id
+  end
+end
--- a/reporting-app/db/schema.rb
+++ b/reporting-app/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2026_06_12_120000) do
+ActiveRecord::Schema[8.0].define(version: 2026_06_16_170507) do
   # These are extensions that must be enabled in order to support this database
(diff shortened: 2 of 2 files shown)


reporting-app/db/migrate/20260616170507_create_denial_response_application_forms.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class CreateDenialResponseApplicationForms < ActiveRecord::Migration[8.0]
+  def change
+    create_table :denial_response_application_forms, id: :uuid do |t|
+      t.uuid :user_id
+      t.integer :status
+      t.datetime :submitted_at
+      t.uuid :certification_case_id
+      t.text :comment
```
giverm, line 10: I know it's different from the existing application form tables, but I'm wondering if we should tighten the schema a bit and not allow `null` for `user_id`, `status`, and `certification_case_id`.

```suggestion
      t.uuid :user_id, null: false
      t.integer :status, null: false, default: 0
      t.datetime :submitted_at
      t.uuid :certification_case_id, null: false
      t.text :comment
```

giverm (approved):

reporting-app/db/migrate/20260616170507_create_denial_response_application_forms.rb
jdettmannnava: I would rather stay consistent. Also, I am not a big fan of database constraints double-enforcing application logic.

Issue #681, opened 2026-06-17
Member dashboard shows "Placeholder" as the exemption type for staff-approved exemptions


giverm:
## Background
When staff approve an exemption request, the member dashboard's exemption history
renders the type as literally **"Placeholder"** instead of the real type (e.g.
"Incarceration"). Reproduced locally.

**Root cause** — a producer/consumer gap:

- `CertificationCase#accept_exemption_request` (`app/models/certification_case.rb:109`)
  records the `:exempt` determination with a hard-coded stub:
  `determination_data: { exemption_type: "placeholder" }`. The method signature is
  `accept_exemption_request(user)` and has no access to the approved form, so it can't
  record the real type.
- The consumer (`MemberDashboardCompliance#exemption_type_label`,
  `member_dashboard_compliance.rb:270`) does
  `I18n.t("exemption_types.#{key}.title", default: key.humanize)`. There is no
  `exemption_types.placeholder` key, so it humanizes to "Placeholder". The reasons
  fallback doesn't rescue this because the recorded reason `exemption_request_compliant`
  doesn't end in `_exempt`.

## Acceptance Criteria
- [ ] Staff-approved exemptions record the real `exemption_type` in the determination,
      sourced from the approved `ExemptionApplicationForm`.
- [ ] Member dashboard renders the real exemption label (en + es-US), not "Placeholder".
- [ ] Specs cover the model recording the real type and the dashboard rendering it.

## Out-of-scope
- Data backfill. Verified on AWS dev that zero determinations carry the "placeholder"
  value (the manual approval path has produced no persisted rows); no production exists.

## Dependencies & Risks
- Pairs with the automated-exemption 500 issue (#680). The fix threads the approved form
  through `accept_exemption_request(user, application_form)` (matching the existing
  `accept_activity_report(user, application_form)` sibling) and passes
  `@task.application_form` from `ReviewExemptionClaimTasksController#update`
  (`app/controllers/review_exemption_claim_tasks_controller.rb:8`), which touches the
  existing callers/specs.

PR #683, opened 2026-06-17
Member dashboard 500s for automated exemption determinations

### LINKED ISSUE #680: Member dashboard 500s for automated exemption determinations
## Background
The member dashboard throws a 500 (`NoMethodError`) for any member whose
certification has an **automated** exemption determination. Reproduced locally;
confirmed on AWS dev that all 1,024 existing exempt determinations are affected.

**Root cause** — a producer/consumer type mismatch on `determination_data`:

- `determination_data` is a `jsonb` column (`db/schema.rb:339`), but
  `CertificationCase#record_exemption_determination`
  (`app/models/certification_case.rb:156`) writes `eligibility_fact.reasons.to_json`
  — a JSON **string** — into it. Postgres stores that as a string scalar, so
  ActiveRecord reads it back as a `String`, not a `Hash`. (It's the only one of the
  four `record_determination!` call sites that calls `.to_json`; the others write Hashes.)
- The dashboard consumer assumes a Hash, with no guard:
  ```ruby
  # app/models/member_dashboard_compliance.rb:294
  def exemption_type_key_from_determination_data(data)
    return nil if data.blank?
    data.stringify_keys["exemption_type"].presence  # String has no #stringify_keys → NoMethodError
  end
  ```
  `build_exemption_history` iterates every exempt determination, so one String row
  500s the page. The same Hash assumption exists at `app/models/determination.rb:115`
  (`ce_calculation_type`).

Shipped in the OSCER-480 exemption-dashboard series (consumer added in #569 / OSCER-409);
adjacent to the slice-3 work in flight (#641).

## Acceptance Criteria
- [ ] Member dashboard renders (no 500) for a member with an automated exemption determination.
- [ ] Consumer guards against a non-Hash `determination_data` in both
      `MemberDashboardCompliance#exemption_type_key_from_determination_data` and
      `Determination#ce_calculation_type`.
- [ ] Automated exemptions display the correct exemption label via the existing
      reasons fallback (their reason codes end in `_exempt`).
- [ ] `record_exemption_determination` writes a Hash (`{}`) to `determination_data`,
      not a JSON string.
- [ ] Specs cover the consumer guard and the producer writing a Hash.

## Out-of-scope
- Data backfill of the existing dev rows. No production environment exists, and the
  consumer guard + reasons fallback render the existing rows correctly without migration.

## Dependencies & Risks
- Pairs with the related "Placeholder" exemption-type bug (separate issue); both touch
  `certification_case.rb` determination writes and the same dashboard consumer.

giverm:
## Ticket

Resolves #680

## Changes

- Stop `CertificationCase#record_exemption_determination` from writing `eligibility_fact.reasons.to_json` (a JSON String) into the jsonb `determination_data` column; it now stores a Hash (`{ "exemption_reasons" => reason_codes }`).
- Rewrite `MemberDashboardCompliance#exemption_history_type_for_determination` to resolve the exemption type by `decision_method`: automated determinations read the type from their reason code, manual determinations read `determination_data` behind an `is_a?(Hash)` guard.
- Add the missing `veteran_disability_exempt` key to `config/locales/services/en.yml` reason codes.
- Add specs covering the crash regression (legacy String `determination_data`), the non-Hash guard, the veteran-disability label, and the producer storing a Hash.

## Context for reviewers

The member dashboard 500s for any member whose certification has an automated exemption determination. `determination_data` is a `jsonb` column, but the automated exemption path serialized its payload with `.to_json` first, so Postgres stored a JSON string scalar and ActiveRecord read it back as a `String`. The dashboard's exemption-history read model assumed a `Hash` (`determination_data.stringify_keys[...]`) and raised `NoMethodError`. On the dev environment all 1,024 exempt determinations are stored this way.

Two design points worth knowing:

- **Why a Hash with reason codes, not `{}`.** `Strata::Determination` validates `determination_data` presence, and Rails treats `{}` as blank, so an empty hash is rejected. Across OSCER, `determination_data` is used as a structured Hash (CE determinations store `HoursBasedDeterminationData#to_h`, the manual exemption stores `{ exemption_type: ... }`), so the automated path now follows that convention by recording the granting reason codes.
- **Why branch on `decision_method`.** Automated exemptions carry their type in the reason code; manual (staff-reviewed) exemptions carry it in `determination_data`. Branching reads each from its authoritative source, and because the automated branch never touches `determination_data`, the existing String rows can no longer crash the page (they label from the reason code instead). The `is_a?(Hash)` guard on the manual branch keeps a malformed row from raising.

The staff-approved "Placeholder" exemption type is a separate, smaller defect tracked in #681. This PR intentionally leaves the manual path unchanged.

## Testing

- Reproduced the bug locally: created and approved an exemption, then confirmed an automated exemption determination 500s the dashboard before the fix.
- TDD: the red run reproduced the exact production failure (`NoMethodError: undefined method 'stringify_keys' for an instance of String`) and the double-encoded payload; green after the fix.
- Verified in the UI after the fix: the member dashboard renders the exemption history for an automated exemption determination without error.
- `member_dashboard_compliance_service_spec.rb`, `certification_case_spec.rb`, `determination_spec.rb`, `certification_business_process_spec.rb`: 101 examples, 0 failures.
- Related suites (dashboard views, compliance/dashboard helpers, `member_status_service`, review-exemption request + model specs): 149 examples, 0 failures.
- Full suite green. RuboCop clean.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/certification_case.rb
+++ b/reporting-app/app/models/certification_case.rb
@@ -149,11 +149,18 @@ class CertificationCase < Strata::Case
       self.exemption_request_approval_status_updated_at = Time.current
       close!
 
+      reason_codes = Determination.to_reason_codes(eligibility_fact)
       certification.record_determination!(
         decision_method: :automated,
-        reasons: Determination.to_reason_codes(eligibility_fact),
+        reasons: reason_codes,
         outcome: :exempt,
-        determination_data: eligibility_fact.reasons.to_json,
+        # determination_data is a jsonb column holding a structured Hash (OSCER convention; see
+        # the Determination docs). For automated exemptions the type is carried by the reason
+        # codes, which the dashboard reads via decision_method branching, so we record those
+        # codes here. It must stay a Hash: writing a JSON String (e.g. reasons.to_json)
+        # double-encodes into the jsonb column and reads back as a String, which 500'd the
+        # member dashboard. See https://github.com/navapbc/oscer/issues/680.
+        determination_data: { "exemption_reasons" => reason_codes },
         determined_at: certification.certification_requirements.certification_date,
         actor:
       )
--- a/reporting-app/app/models/member_dashboard_compliance.rb
+++ b/reporting-app/app/models/member_dashboard_compliance.rb
@@ -274,26 +274,41 @@ class MemberDashboardCompliance
     I18n.t("exemption_types.#{key}.title", default: key.to_s.humanize)
   end
 
+  # The exemption type lives in different places depending on how the exemption was decided.
+  # Automated (eligibility-derived) exemptions encode the type in their reason codes, while
+  # manual (staff-reviewed) exemptions record the member's claimed type in
+  # +determination_data["exemption_type"]+. Branch on +decision_method+ so each reads its
+  # authoritative source — this also keeps a legacy/malformed +determination_data+ on an
+  # automated row from ever reaching the Hash read (see #680).
   def exemption_history_type_for_determination(det)
-    type_key = exemption_type_key_from_determination_data(det.determination_data)
-    return [ type_key, exemption_type_label(type_key) ] if type_key.present?
-
-    exempt_reason = Array(det.reasons).find { |r| r.to_s.end_with?("_exempt") }
-    if exempt_reason
-      label = I18n.t(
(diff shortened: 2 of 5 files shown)


jdettmannnava (approved):
LGTM!
I agree on the guard over backfill approach

PR #685, opened 2026-06-17
Member dashboard shows "Placeholder" as the exemption type for staff-approved exemptions

### LINKED ISSUE #681: Member dashboard shows "Placeholder" as the exemption type for staff-approved exemptions

giverm:
## Ticket

Resolves #681

## Changes

- Change `CertificationCase#accept_exemption_request(user)` to `accept_exemption_request(user, application_form)` and record the form's real `exemption_type` in `determination_data` instead of the hard-coded `"placeholder"` stub.
- Pass `@task.application_form` from `ReviewExemptionClaimTasksController#update`.
- Update the existing call sites in the model, business-process, exemption-form, and dashboard-view specs to pass the reviewed form.
- Add specs: the model records the form's type (with an `incarceration` case proving it is sourced from the form), the controller flow records it (request spec), and the dashboard renders the real type label (view spec).

## Context for reviewers

When staff approved an exemption, `accept_exemption_request` recorded `determination_data: { exemption_type: "placeholder" }` because the method had no access to the reviewed form. The member dashboard reads that value back, so approved exemptions displayed their type as the literal "Placeholder".

The fix mirrors the existing `accept_activity_report(user, application_form)` sibling: the controller already holds the review task's bound `application_form`, so it threads that through and the determination records `application_form.exemption_type` as a snapshot at approval time. This is the OSCER convention for `determination_data` (a structured Hash describing the determination), consistent with how the hours/income paths use the column.

Out of scope, intentionally: data backfill (verified on dev that zero determinations carry the "placeholder" value; no production). Pairs with #683 (the automated-exemption 500 fix, #680); the two touch different methods in `certification_case.rb` and do not conflict.

## Testing

- TDD: red run showed `determination_data` recording `"placeholder"` against the expected form type; green after threading the form through.
- Reproduced and verified in the UI: a staff-approved exemption now shows its real type on the member dashboard instead of "Placeholder".
- `certification_case_spec.rb`, `review_exemption_claim_tasks_spec.rb` (request), and `index.html.erb_spec.rb` (view) cover the model recording, the controller flow, and the rendered label.
- Full suite green (2250 examples, 0 failures); dashboard view spec green including the added render assertion. RuboCop clean.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/controllers/review_exemption_claim_tasks_controller.rb
+++ b/reporting-app/app/controllers/review_exemption_claim_tasks_controller.rb
@@ -5,7 +5,7 @@ class ReviewExemptionClaimTasksController < TasksController
     kase = @task.case
 
     if approving?
-      kase.accept_exemption_request(current_user)
+      kase.accept_exemption_request(current_user, @task.application_form)
       @task.approval_status = :approved
       notice = t("tasks.details.approved_message")
     elsif denying?
--- a/reporting-app/app/models/certification_case.rb
+++ b/reporting-app/app/models/certification_case.rb
@@ -99,7 +99,7 @@ class CertificationCase < Strata::Case
     Strata::EventManager.publish(event_name, { case_id: id, certification_id: certification_id, application_form_id: application_form.id })
   end
 
-  def accept_exemption_request(user)
+  def accept_exemption_request(user, application_form)
     transaction do
       self.exemption_request_approval_status = "approved"
       self.exemption_request_approval_status_updated_at = Time.current
@@ -112,7 +112,7 @@ class CertificationCase < Strata::Case
         outcome: :exempt,
         determined_at: certification.certification_requirements.certification_date,
         determination_data: {
-          exemption_type: "placeholder"
+          exemption_type: application_form.exemption_type
         },
         actor: user
       )
--- a/reporting-app/spec/business_processes/certification_business_process_spec.rb
+++ b/reporting-app/spec/business_processes/certification_business_process_spec.rb
@@ -203,7 +203,7 @@ RSpec.describe CertificationBusinessProcess, type: :business_process do
       expect(certification_case).to be_open
 
       # Step 3: Staff approves exemption
-      certification_case.accept_exemption_request(user)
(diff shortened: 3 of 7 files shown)


jdettmannnava (approved):
LGTM!

Issue #666, opened 2026-06-15
Async CE status via job + polling: OSCER to state system return contract


sumiat:
## Summary 
A state system will call OSCER at MAGI wrap-up (expansion cases); OSCER accepts the request, returns a jobID, and the state polls the jobID for status until OSCER returns a final CE determination or an error. This ticket covers the core OSCER work to build that async job flow: accept the request, run the determination on the electronic data sources (in another piece of work), expose job status on poll, return the CE determination payload (status, reason, source, timestamp), handle duplicate inbound calls, and define the JSON request/response objects OSCER owns. Response is JSON. State-side handling of what it receives is out of scope.

## Gherkin User Stories
Scenario: OSCER accepts the request and returns a job to poll
  Given the state calls OSCER at MAGI wrap-up for an expansion case
  When OSCER receives the request
  Then OSCER returns a jobID synchronously
  And the jobID is associated with the originating case/correlation ID
  And the initial job status is "waiting_for_response"

Scenario: State polls a job still in progress
  Given OSCER has a job in progress for a case
  When the state polls the jobID
  Then OSCER returns status "waiting_for_response"

Scenario: State polls a job that has a determination
  Given OSCER has computed a final CE determination for a job
  When the state polls the jobID
  Then OSCER returns status from the locked CE status enum
  And the response includes a reason value(s)
  And the response includes the data source
  And the response includes a determination timestamp
  And the response echoes the originating case/correlation ID

Scenario: A determination stalls waiting on a data source
  Given OSCER must query a source not available at call time
  When the source has not returned after the defined time limit
  Then state system will poll for the status update

Scenario: State polls a failed job
  Given a determination job has failed
  When the state polls the jobID
  Then OSCER returns an error status for that job

## Acceptance Criteria
Include activities or user scenarios/stories that must be completed to finish the story. 

- [ ] On receipt, OSCER returns a jobID synchronously, associated with the originating case/correlation ID, with initial status waiting_for_response.
- [ ] State retrieves status by polling the jobID; OSCER does not push a callback.
- [ ] A polled completed job returns **status**, **reason**, **source**, **timestamp**, and the **case/correlationID**
- [ ] **reason** and **source** are free text
- [ ] **timestamp** is the determination time
- [ ] A job stalled on an unavailable data source returns waiting_for_response on poll, returns the final determination when ready, and stops issuing updates after the defined time limit.
- [ ] A failed job returns an error status on poll. 
[...]

PR #691, opened 2026-06-22
OSCER-666: return latest determination outcome in api response

jdettmannnava:
## Ticket

Stage 1 of #666 

## Changes

- Api::Certifications::Response now has `outcome` attribute
- Added Api::Certifications::Outcome value object
- openapi doc updated

## Context for reviewers

This is step 1 in providing an async endpoint to clients wanting complete external checks for community engagement. 

The Certification API already has a synchronous `create` endpoint that already kicks off the external checks. Adding an asynchronous response to the endpoint seems to be the most straightforward way to implement the functionality of #666. By adding the outcome needed by #666 to all Certification API responses, we maintain the integrity of the API, as we will return the same response format to both the synchronous and asynchronous API calls to create a Certification.

Outcome attributes are based on needs of Alaska, which seem reasonable enough to make the default.

## Testing

- New functionality, so regressions unlikely
- Passed additional tests
- Verified appropriate responses when calling Api::CertificationsController#show

♻️ Environment destroyed ♻️


reporting-app/app/models/certification.rb
```diff
@@ -43,6 +43,11 @@ def member_email
     member_account_email || member_contact_email
   end
 
+  def outcome
+    @outcome ||= Api::Certifications::Outcome.from_certification(self)
+    @outcome
```
giverm: I believe line 47 will return the outcome object, so this line can be removed.

```suggestion
```

reporting-app/app/models/api/certifications/outcome.rb
```diff
@@ -0,0 +1,25 @@
...
+    obj = new(timestamp: determination.created_at)
+    if determination.not_compliant? && determination.automated?
+        obj.status = "indeterminate"
+    elsif determination.not_compliant?
+        obj.status = "not_compliant"
+    else
+        obj.status = determination.outcome
+        obj.reason = determination.reasons.first
+        obj.source = determination.automated? ? "api" : "member"
+    end
```
giverm, line 22: Extra indentation 

```suggestion
    if determination.not_compliant? && determination.automated?
      obj.status = "indeterminate"
    elsif determination.not_compliant?
      obj.status = "not_compliant"
    else
      obj.status = determination.outcome
      obj.reason = determination.reasons.first
      obj.source = determination.automated? ? "api" : "member"
    end
```

giverm (approved):
Looks great!

reporting-app/lib/assets/oas.json
```diff
@@ -491,6 +518,12 @@
               { "type": "null" }
             ]
           },
+	  "outcome": {
+            "anyOf": [
+              { "$ref": "#/components/schemas/CertificationOutcome" },
+              { "type": "null" }
+            ]
+	  },
```
tomas-nava:
```suggestion
          "outcome": {
                  "anyOf": [
                    { "$ref": "#/components/schemas/CertificationOutcome" },
                    { "type": "null" }
                  ]
          },
```
fix indentation

reporting-app/lib/assets/oas.json
```diff
@@ -450,6 +450,33 @@
...
+          "source": {
+            "type": ["string", "null"],
+            "description": "Which source provided the reason"
+          },
+          "timestamp": {
+            "type": "string",
+            "format": "date-time",
+            "description": "When the status was set"
+          }
+	},
```
tomas-nava:
```suggestion
        "type": "object",
          "properties": {
            "status": {
              "type": "string",
              "enum": ["compliant", "exempt", "indeterminate", "not_compliant"],
              "description": "Current status of community engagement requirement checks"
            },
            "reason": {
              "type": ["string", "null"],
              "description": "Which exemption or community engagement requirement led to the decision"
            },
            "source": {
              "type": ["string", "null"],
              "description": "Which source provided the reason"
            },
            "timestamp": {
              "type": "string",
              "format": "date-time",
              "description": "When the status was set"
            }
        },
```
fix indent

reporting-app/lib/assets/oas.json
```diff
@@ -450,6 +450,33 @@
           }
         ]
       },
```
tomas-nava, line 452: You could add examples of the new field to one or more of the objects in the `examples` list that starts at line 41

reporting-app/app/models/api/certifications/outcome.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class Api::Certifications::Outcome < ValueObject
+  attribute :status, :string
+  attribute :reason, :string
+  attribute :source, :string
+  attribute :timestamp, :datetime
+
+  def self.from_certification(certification)
+    determination = Determination.where(subject: certification).first
```
tomas-nava: Speaking from ignorance; is there a circumstance in which we wouldn't just want the first `Determination`? Might we want the most recent? Or prefer one that has a `"compliant"` outcome? Or just return all of them?

reporting-app/spec/models/certification_spec.rb
```diff
@@ -149,4 +149,90 @@
       expect(certification.origin).to eq(origin)
     end
   end
+
```
tomas-nava, line 152: Should there also be new tests in `spec/requests/api/certifications_spec.rb`?

reporting-app/app/models/api/certifications/outcome.rb
nkutub: I agree, I have been bitten by this. we should at least order by datatime to make sure the first one is the latest one.

jdettmannnava: This was an artifact of testing. It should be the most recent. I was trying to get it to fail...

reporting-app/lib/assets/oas.json
jdettmannnava: Added to response (line 588)

reporting-app/spec/models/certification_spec.rb
jdettmannnava: They sort of are - `spec/requests/api/certifications_spec.rb` checks that the outcome matches the openapi spec, and they fail if, for example, the timestamp or status is not present.

reporting-app/app/models/api/certifications/outcome.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class Api::Certifications::Outcome < ValueObject
+  attribute :status, :string
+  attribute :reason, :string
+  attribute :source, :string
+  attribute :timestamp, :datetime
+
+  def self.from_certification(certification)
+    determination = Determination.where(subject: certification).order(created_at: :desc).first
```
tomas-nava:
```suggestion
    determination = Determination.where(subject: certification).latest_first.first
```
What you have is fine, but `Determinable` does have this built-in `latest_first` helper method

tomas-nava (approved):
Looks good! I made one suggestion you can ignore.

PR #703, opened 2026-06-24
Allow deployer-defined feature flags via config/custom YAML

### LINKED ISSUE #607: Allow deployer-defined feature flags via config/custom YAML
## Allow deployer-defined feature flags via config/custom YAML

> **Repurposed 2026-06-22** (OSS epic [#527](https://github.com/navapbc/oscer/issues/527) wound down). The mechanism is unchanged — extend the `config/custom/*.yml` deep-merge pattern (the **Config** mechanism in `CUSTOMIZATION.md`, the one `exemption_types.yml` already uses) to feature flags. Dropped: the `config/<deployment_namespace_snake>/` rename and the Copier `_skip_if_exists` / `nava-platform app update` framing (those were template-distribution concerns; the namespace rename never landed and `config/custom/` is the real path on `main`). Parked forward work.

## Problem

OSCER's feature-flag mechanism (`config/initializers/feature_flags.rb`) reads env vars at runtime, but the **flag registry** (`Features::FEATURE_FLAGS` hash) lives in Ruby. A downstream deployment can toggle existing flags via its environment, but cannot define its own deployment-specific flags without editing that Ruby file — which puts deployment-specific code in a file that conflicts on every upstream sync.

This is the same asymmetry that `config/custom/exemption_types.yml` already solves for exemption types: a Ruby-defined registry deep-merged with an optional deployment-owned YAML override that lives in an override seam (so it survives upstream syncs cleanly). Feature flags should follow the same shape — minus the regulatory weight; feature flags are OSCER-shipped built-ins, not federal-floor minimums.

## Proposed approach

Add a new Config override file: `config/custom/feature_flags.yml`, mirroring `config/custom/exemption_types.yml`.

```yaml
# Example: config/custom/feature_flags.yml
arkansas_report:
  env_var: FEATURE_ARKANSAS_REPORT
  default: false
  description: Enable Arkansas-specific compliance report
```

Loader behavior:
1. Frozen `Features::FEATURE_FLAGS` Ruby constant keeps OSCER-shipped built-ins (today: `doc_ai`); every deployment inherits these.
2. Optional `config/custom/feature_flags.yml` adds deployment-defined flags.
3. Deep-merge at boot, **additive only** — deployment entries add flags; they cannot remove or disable OSCER-shipped flags (use the env var for that).
4. Validation: each entry needs `env_var` (matching `FEATURE_<NAME>` shape) and a boolean `default`. Malformed entries fail loudly at boot.
5. Auto-generated `Features.<name>_enabled?`, `Features.enabled?(:<name>)`, and test helpers (`with_<name>_enabled` / `with_<name>_disabled`) work for both sets.

## Acceptance criteria

- [ ] New Config override `config/custom/feature_flags.yml`, documented alongside `exemption_types.yml`.
- [ ] Optional file — missing or empty means "no deployment-defined flags."
- [ ] Deep-merge: deployment entries are added to the in-memory registry; `Features.all_flags` returns both sets unioned.
- [ ] Auto-generated methods + test helpers work for deployment-defined flags.
- [ ] Validation errors at boot are loud and specific (which entry, which field).
- [ ] Spec covers: happy path, missing file, empty file, malformed entry (missing keys, wrong types), duplicate-name collision between OSCER-shipped and deployment-defined.

## Out of scope

- Removing/disabling OSCER-shipped flags from deployment YAML (use the env var).
- Per-environment overrides via YAML (env vars cover that).
- UI for managing flags.

## Cross-refs

- Was part of [#527](https://github.com/navapbc/oscer/issues/527) (epic backlogged).
- Builds on the `config/custom/` deep-merge infrastructure from #540 (PR #564).
- Companion to `reporting-app/CUSTOMIZATION.md` (now on `main`, originated in [#539](https://github.com/navapbc/oscer/issues/539)) — the deployment-facing doc for flag customization lives there.

giverm:
## Ticket

Resolves #607

## Changes

- Add `FeatureFlagsLoader` service that reads an optional `config/custom/feature_flags.yml`, validates each entry, and unions deployment-defined flags onto the OSCER-shipped built-ins.
- Build a merged `Features::REGISTRY` at boot (built-ins ∪ deployment flags). `FEATURE_FLAGS` stays the frozen built-ins-only constant; `*_enabled?` generation, `enabled?`, `all_flags`, and `flag_config` now read `REGISTRY`.
- Generate `with_<name>_enabled` / `with_<name>_disabled` test helpers over the merged registry, so deployment-defined flags get helpers too.
- Ship `config/custom/feature_flags.yml` as an empty override (`{}`) with a documented header, mirroring `config/custom/exemption_types.yml`.
- Document deployment-defined flags in `CUSTOMIZATION.md` (Config section plus a quick-reference row).
- Validate at boot, failing loudly and naming the offending entry and field: require an `env_var` matching `FEATURE_<NAME>` and a boolean `default`; reject unknown keys; reject a flag that collides with a built-in by name or by `env_var`, and reject duplicate `env_var`s across deployment entries.

## Context for reviewers

The flag registry lives in Ruby (`Features::FEATURE_FLAGS`), so a deployment can toggle existing flags via environment variables but cannot define its own without editing an OSCER-owned file that conflicts on every upstream sync. This extends the same `config/custom/` deep-merge override pattern that `exemption_types.yml` already uses, applied to feature flags.

A few decisions worth knowing while reading:

- **Additive only, intentionally not `deep_merge`.** Unlike `ExemptionTypesLoader` (where the deployment owns the data and may override defaults), feature flags are OSCER-owned rollout state, so the loader only *adds* deployment flags and raises on any collision. A header comment in `feature_flags_loader.rb` documents this so the two loaders aren't "aligned" by a later refactor.
- **A flag's behavioral identity is its `env_var`, not its name.** The collision guard rejects a deployment flag whose `env_var` matches a built-in's even under a different name, which is what actually enforces the "deployments cannot disable a built-in" rule. Toggling or disabling a built-in stays an environment-variable operation.
- **Load order.** `REGISTRY` is built before the metaprogramming loops that generate the `*_enabled?` methods and the test helpers. The loader is `require`d explicitly in the initializer because Zeitwerk autoloading is not available during initializers, matching `config/initializers/exemption_types.rb`.
- **Out of scope** (per the issue): disabling built-ins from YAML, per-environment YAML overrides, and any flag-management UI.

## Testing

- RuboCop clean on all changed files.
- Targeted specs pass (41 examples, 0 failures): `feature_flags_loader_spec`, `features_spec`, `feature_flag_helpers_spec`, `feature_flag_helper_spec`. Coverage spans happy path, missing/empty/comments-only file, malformed YAML, non-Hash top level, disallowed Ruby tag, missing and malformed `env_var`, missing and non-boolean `default`, built-in name collision, `env_var` collision (built-in reuse and two deployment entries), unknown key, and generated-method/helper behavior for a registered flag.
- The full suite was not run yet; it should be run before merge, since targeted runs trip the global SimpleCov coverage gate by design.

♻️ Environment destroyed ♻️


reporting-app/app/services/feature_flags_loader.rb
```diff
@@ -0,0 +1,130 @@
...
+# deep_merges deployment values over OSCER defaults (the deployment owns the
+# data and may override it), whereas feature flags are OSCER-owned rollout
+# state, so this loader is purely ADDITIVE and raises on any collision. Do not
+# "align" the two by introducing deep_merge here.
+module FeatureFlagsLoader
+  class ConfigurationError < StandardError; end
+
+  # Deployment-defined env vars must match the same FEATURE_<NAME> convention
+  # the built-ins follow: "FEATURE_" + uppercase letters/digits/underscores.
+  ENV_VAR_PATTERN = /\AFEATURE_[A-Z0-9_]+\z/
```
jgavin-nava, line 30: WE validate the env_var_pattern do we also want to validate the FLAG_NAME to be snake case?

jgavin-nava (approved):
LGTM otherwise!!

PR #711, opened 2026-06-24
OSCER-666 async endpoint

### LINKED ISSUE #666: Async CE status via job + polling: OSCER to state system return contract

jdettmannnava:
## Ticket

Resolves #666

## Changes

- `outcome` endpoint added to Api::CertificationsController
- `create` endpoint returns `accepted` response when `prefer: respond-async` header in request to Api::CertificationsController
- OpenApi documentation updated

## Context for reviewers

The processing for determinations based on external data is not really a single job, so it felt cleaner to create a new endpoint that only responds definitively once there is an outcome/determination.

Error handling remains primitive/non-existent. Can be added as follow-up ticket.

## Testing

Tests added and `make build lint test` passes.

Tested create with `prefer: respond-async` manually and worked as expected.

♻️ Environment destroyed ♻️


giverm (approved):
LGTM!

PR #739, opened 2026-07-02
Gate demo certifications tooling behind FEATURE_DEMO_CERTIFICATIONS (OSCER-737)

### LINKED ISSUE #737: Gate demo certifications tooling behind FEATURE_DEMO_CERTIFICATIONS (default off)
## Background

The demo certifications UI lets anyone create **real** `Certification` records via the production `Certifications::CreationService`. It is intended as internal seeding / demo tooling, not a member-facing product feature.

Today:

- Routes are mounted unconditionally in `reporting-app/config/routes.rb` (`GET /demo`, `GET/POST /demo/certifications`).
- `Demo::CertificationsController` has no `authenticate_user!`, skips Pundit (`skip_after_action :verify_authorized`), and has no environment guard.
- `DemoController` (demo index) inherits from `ActionController::Base` with no auth or gating.
- Contrast: `DevController` redirects outside `Rails.env.development?` (`reporting-app/app/controllers/dev_controller.rb`).

Request specs in `reporting-app/spec/requests/demo/certifications_spec.rb` currently assume unauthenticated access works.

OSCer already has a feature-flag registry (`reporting-app/config/initializers/feature_flags.rb`, `Features` module, env vars like `FEATURE_DOC_AI`, Terraform via `infra/…/app-config/feature_flags.tf`). See `docs/feature-flags.md`.

**Team direction:** Put demo certifications behind a feature flag (default off), rather than leaving it open in every deployed environment.

## User Story

As an **OSCer maintainer**, I want demo certification creation **disabled by default** in deployed environments and **explicitly enableable** only where we need it (e.g. local dev, a staging UAT env), so we do not expose an unauthenticated write path that inserts PII into the production datastore.

### Feature flag registration

- [ ] Add a built-in flag to `Features::FEATURE_FLAGS`, e.g. `demo_certifications`:
  - `env_var: "FEATURE_DEMO_CERTIFICATIONS"`
  - `default: false`
  - Short description: enable demo certification seeding UI
- [ ] Boot exposes `Features.demo_certifications_enabled?` (and existing test helpers via `FeatureFlagHelpers`).

### Application gating (fail closed)

- [ ] When the flag is **disabled** and not in `development`/`test`:
  - `GET /demo`, `GET /demo/certifications/new`, and `POST /demo/certifications` return **404** (preferred) or redirect to root — consistent behavior across demo controllers.
- [ ] When the flag is **enabled** **or** `Rails.env.development?` / `Rails.env.test?`, current demo behavior is unchanged (no auth required in dev/test unless we add staff auth in a follow-up).
- [ ] Gate applies to **`DemoController`** and **`Demo::CertificationsController`** (and any other demo routes added later).

### Implementation notes (suggested)

- Controller `before_action :ensure_demo_certifications_enabled` (shared concern or base class) checking `Features.demo_certifications_enabled? || Rails.env.development? || Rails.env.test?`.
- Optionally wrap route mounting in `config/routes.rb` with the same condition so disabled routes are not even registered in production (defense in depth).
- Document `FEATURE_DEMO_CERTIFICATIONS=true` in `reporting-app/local.env.example` for developers who run with `RAILS_ENV=development` but want to test the flag explicitly.

### Infrastructure

- [ ] Register the flag in Terraform (`infra/…/app-config/feature_flags.tf`) with default **`false`** for all environments.
- [ ] Document how to enable for a specific non-prod env (e.g. staging UAT) via `feature_flag_overrides` in that env's config — **production must remain false**.

### Tests

- [ ] Update `spec/requests/demo/certifications_spec.rb`:
  - Existing examples run with flag enabled (or in test env where gate allows access).
  - New examples: with flag disabled and `RAILS_ENV` simulated as production (or equivalent), `GET /demo/certifications/new` and `POST /demo/certifications` are **not** successful (404/redirect).
- [ ] Add unit/request coverage for `Features.demo_certifications_enabled?` if not covered by existing flag patterns.

## Out-of-scope (follow-up tickets)

- Requiring **staff authentication** when the flag is enabled outside `development` (recommended hardening for shared staging; not required for this flag-only slice).
- Fixing unrelated demo security items (email-based member auth, cross-member case IDOR, etc.).
- Removing demo tooling entirely — this ticket only adds gating.

## Dependencies & Risks

- **Risk:** Flag mis-set to `true` in production re-opens the vulnerability — mitigate with Terraform default `false`, code default `false`, and a request spec that asserts prod-like envs are blocked.
- **Dependency:** Terraform apply / deploy for env-specific enablement in staging (ops step after merge).

## Open questions

- Should **`development`** auto-allow demo without the flag (match today's DX), or require `FEATURE_DEMO_CERTIFICATIONS=true` locally too?
- Do we need demo enabled in **any** shared staging env today? If yes, which env name gets the Terraform override?

## Links & Resources

- `reporting-app/app/controllers/demo/certifications_controller.rb`
- `reporting-app/app/controllers/demo_controller.rb`
- `reporting-app/config/routes.rb` (demo routes ~194–198)
- `reporting-app/config/initializers/feature_flags.rb`
- `reporting-app/app/controllers/dev_controller.rb` (env-gate precedent)
- `docs/feature-flags.md`
- `reporting-app/spec/requests/demo/certifications_spec.rb`

jgavin-nava:
## Ticket

Resolves [OSCER-737](https://github.com/navapbc/oscer/issues/737)

## Changes

- Added built-in `demo_certifications` flag to `Features::FEATURE_FLAGS` (`FEATURE_DEMO_CERTIFICATIONS`, default off).
- Introduced `DemoAccessGate` with shared `access_allowed?` predicate used by controller `before_action` and route constraints.
- Added `Demo::BaseController` so namespace demo controllers inherit the gate automatically.
- Wrapped demo routes in `config/routes.rb` with a constraint mirroring `DemoAccessGate` (defense in depth).
- Set `FEATURE_DEMO_CERTIFICATIONS = "false"` in all Terraform env configs via `service_override_extra_environment_variables`.
- Documented the flag in `local.env.example` and clarified in `feature_flags.tf` that OSCER app flags use `FEATURE_*` env vars, not the legacy `FF_*` Parameter Store map.
- Added request specs for feature gating and registry coverage for the new flag.

## Context for reviewers

**Infra wiring:** `FEATURE_DEMO_CERTIFICATIONS` is set through each environment's `service_override_extra_environment_variables` block (same pattern as `FEATURE_DOC_AI`), not through `infra/reporting-app/app-config/env-config/feature_flags.tf` / `FF_*` Parameter Store entries. The Rails `Features` module reads `FEATURE_*` environment variables directly.

**Local DX:** In `development` and `test` (`Rails.env.local?`), demo routes remain available regardless of the flag so developer and CI behavior is unchanged.

**Staging UAT:** To seed certifications in staging, temporarily set `FEATURE_DEMO_CERTIFICATIONS = "true"` in `staging.tf`, deploy, run UAT, then revert to `"false"` and re-deploy.

**Security:** Demo tooling creates real `Certification` records with no authentication. Production must keep the flag `"false"`.

## Testing

From `reporting-app/`:

```bash
make test args='spec/requests/demo/index_spec.rb spec/requests/demo/certifications_spec.rb spec/lib/features_spec.rb'
make lint
```

All targeted specs pass. Feature-gating examples stub `Rails.env` as production and assert 404 when the flag is off, successful access when on.

Made with [Cursor](https://cursor.com)

♻️ Environment destroyed ♻️

### DIFF
--- a/infra/reporting-app/app-config/dev.tf
+++ b/infra/reporting-app/app-config/dev.tf
@@ -38,6 +38,6 @@ module "dev_config" {
     FEATURE_BATCH_UPLOAD_V2     = "true"
     PERFORM_EMAIL_DELIVERY      = "false"
     FEATURE_DOC_AI              = "false"
-    FEATURE_DEMO_CERTIFICATIONS = "false"
+    FEATURE_DEMO_CERTIFICATIONS = "true"
   }
 }
--- a/infra/reporting-app/app-config/dev.tf
+++ b/infra/reporting-app/app-config/dev.tf
@@ -34,9 +34,10 @@ module "dev_config" {
   # }
 
   service_override_extra_environment_variables = {
-    ENABLE_LOOKBOOK         = "true"
-    FEATURE_BATCH_UPLOAD_V2 = "true"
-    PERFORM_EMAIL_DELIVERY  = "false"
-    FEATURE_DOC_AI          = "false"
+    ENABLE_LOOKBOOK             = "true"
+    FEATURE_BATCH_UPLOAD_V2     = "true"
+    PERFORM_EMAIL_DELIVERY      = "false"
+    FEATURE_DOC_AI              = "false"
+    FEATURE_DEMO_CERTIFICATIONS = "false"
   }
 }
--- a/infra/reporting-app/app-config/env-config/feature_flags.tf
+++ b/infra/reporting-app/app-config/env-config/feature_flags.tf
@@ -1,4 +1,10 @@
 locals {
+  # Legacy FF_* Parameter Store flags (see infra/reporting-app/service/feature_flags.tf).
+  # OSCER app feature toggles read FEATURE_* env vars via the Features module
+  # (config/initializers/feature_flags.rb) and are set per environment in each
+  # env's service_override_extra_environment_variables (e.g. FEATURE_DOC_AI,
+  # FEATURE_DEMO_CERTIFICATIONS) — not in this map.
+  #
   # Map from feature flags to their default values (true or false)
   feature_flag_defaults = {
     # Example feature flags
--- a/infra/reporting-app/app-config/env-config/feature_flags.tf
+++ b/infra/reporting-app/app-config/env-config/feature_flags.tf
@@ -1,10 +1,4 @@
 locals {
-  # Legacy FF_* Parameter Store flags (see infra/reporting-app/service/feature_flags.tf).
-  # OSCER app feature toggles read FEATURE_* env vars via the Features module
-  # (config/initializers/feature_flags.rb) and are set per environment in each
-  # env's service_override_extra_environment_variables (e.g. FEATURE_DOC_AI,
-  # FEATURE_DEMO_CERTIFICATIONS) — not in this map.
-  #
   # Map from feature flags to their default values (true or false)
   feature_flag_defaults = {
     # Example feature flags
--- a/infra/reporting-app/app-config/prod.tf
+++ b/infra/reporting-app/app-config/prod.tf
@@ -25,10 +25,4 @@ module "prod_config" {
   # Defaults to `false`. Uncomment the next line to enable.
   # ⚠️ Warning! It is not recommended to enable this in a production environment.
   # enable_command_execution = true
-
-  service_override_extra_environment_variables = {
-    # Demo certification seeding UI. This creates real Certification records via
-    # an unauthenticated write path, so it MUST stay disabled in production.
-    FEATURE_DEMO_CERTIFICATIONS = "false"
-  }
 }
--- a/infra/reporting-app/app-config/prod.tf
+++ b/infra/reporting-app/app-config/prod.tf
@@ -25,4 +25,10 @@ module "prod_config" {
   # Defaults to `false`. Uncomment the next line to enable.
   # ⚠️ Warning! It is not recommended to enable this in a production environment.
   # enable_command_execution = true
+
+  service_override_extra_environment_variables = {
+    # Demo certification seeding UI. This creates real Certification records via
+    # an unauthenticated write path, so it MUST stay disabled in production.
+    FEATURE_DEMO_CERTIFICATIONS = "false"
+  }
 }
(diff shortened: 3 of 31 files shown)


giverm (approved):

giverm (approved):
Approving — clean, well-scoped fail-closed gating. Ran a security/quality/context agent pass plus a freeform review: security and quality came back clean (5/5 each), and all five ACs are functionally met. A few non-blocking notes.

**Should-fix (non-blocking):**
1. The controller `before_action` guard in `DemoAccessGate` (the `head :not_found` branch) isn't exercised today — every current demo route is caught by the `constraints` block first, so requests 404 at route recognition before reaching the controller. It's intentional defense-in-depth for demo routes added outside that block later, but the branch has no coverage. Consider a small controller-level test (anonymous controller including the concern) to lock in the belt-and-suspenders guarantee, or just note it's intentionally uncovered.
2. `local.env.example` sets `FEATURE_DEMO_CERTIFICATIONS=false` while the ticket suggested `=true`. The `=false` is arguably the better choice (matches `FEATURE_DOC_AI=false` right above it and the fail-closed default) — just confirming it's deliberate.

**Nice-to-have:**
- More surgical env stub in the specs: `allow(Rails.env).to receive(:local?).and_return(false)` rather than stubbing `Rails.env` app-wide.
- Optional direct assertion on `Features.demo_certifications_enabled?` (currently covered transitively via the gate spec).

**On the infra AC — good call:** using `service_override_extra_environment_variables` in `dev.tf` rather than the `feature_flags.tf` / `feature_flag_defaults` mechanism the ticket referenced was correct. That template-infra mechanism is unwired in OSCER (empty `feature_flag_defaults`, and it emits `FF_*` container secrets while the `Features` module reads `FEATURE_*`), so following the AC literally would have been a silent no-op. You matched the pattern every real flag uses. I'll document that template divergence separately so it doesn't trip up the next flag ticket.

Also verified the flag flows end-to-end into the PR environment (`FEATURE_DEMO_CERTIFICATIONS` present on the dev-config task def). Nice work.

PR #743, opened 2026-07-06
OSCER-735: Add exception business process

### LINKED ISSUE #735: External Exception Check Scaffolding
## Background
The certification workflow runs two automated determination checks — an external exclusion check,
then an external community-engagement check — before any member action is required. We are introducing a
third, an external exception check, that will host a set of hard-coded, individually-configurable
checks; when any applies, the case resolves without the member reporting activities.

This story stands up the step and the shared configuration surface the individual checks will plug
into, with no checks implemented yet. The step runs automatically between the exclusion and community-engagement
checks, finds no exception, and passes the case through to the community-engagement check unchanged.
Later stories add the checks.

## User Story
As the team building the certification workflow, we want the external exception check step and its
shared configuration surface in place, so that each individual exception check can be added later as a thin,
independently shippable slice.

## Acceptance Criteria
- [ ] A new external exception check system process runs automatically after the external exclusion
      check and before the external community-engagement check, parallel in architecture to the
      existing determination checks: it records its determination on the case and publishes domain
      events, while the process owns transitions and notifications.
- [ ] With no checks implemented, the step always determines "no exception" and the case continues
      to the community-engagement check unchanged.
- [ ] The positive (exception-applies) path is wired end to end even though it is currently
      unreachable: it records the exception determination on the certification case through a new
      certification-case call (paralleling, but distinct from, the exclusion determination check's
      recording call), publishes the corresponding event, and transitions the case so the member is
      not routed to the community-engagement check or activity reporting. The positive event routes
      to the member's existing `exempt_email` notification, the same one the exclusion outcome uses.
- [ ] The step records its decision in the audit log, consistent with how the existing
      determination services write to the audit log with an 'excepted' outcome.
- [ ] Behavior is verified by automated tests and the Rails console: the step runs, passes through
      as a no-op.

## Out-of-scope
- Configuration of checks
- The individual checks and the member-data inputs they read (later stories).
- Any change to the Certification API or the bulk-upload path.
- Member- or staff-facing production UI.

## Links & Resources
- [Tech Spec](https://docs.google.com/document/d/1NeAEnIwRnBgtwWt_7tonPDdIZYh_KryhvQhHxXGPs6Q/edit?usp=sharing)

jdettmannnava:
## Ticket

Resolves #735 

## Changes
- ExceptionDeterminationService added (with no checks)
- System process added to business process for calling ExceptionDeterminationService
- Plumbing added to CertificationCase to record positive result of ExceptionDeterminationService

## Context for reviewers

We need to check for exceptions as a separate task from checking for exclusions. Exclusions were added before community engagement checks as a matter of convenience. Adding the actual exclusion checks will be a subsequent ticket.

Exempt (from staff-made decisions), excluded, and excepted results all lead to the same email. Fixing that will be made in a later ticket if it is needed.

## Testing
`make lint test` passes.

Generating an unresolved case via demo showed (failed) audit lines made in the correct order.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/determination.rb
+++ b/reporting-app/app/models/determination.rb
@@ -94,7 +94,7 @@ class Determination < Strata::Determination
   VALID_REASONS = REASON_CODE_MAPPING.values.freeze
 
   enum :decision_method, { automated: "automated", manual: "manual" }
-  enum :outcome, { compliant: "compliant", exempt: "exempt", excluded: "excluded", not_compliant: "not_compliant" }
+  enum :outcome, { compliant: "compliant", exempt: "exempt", excluded: "excluded", excepted: "excepted", not_compliant: "not_compliant" }
 
   validates :reasons, presence: true, inclusion: { in: VALID_REASONS }
 
--- a/reporting-app/spec/business_processes/certification_business_process_spec.rb
+++ b/reporting-app/spec/business_processes/certification_business_process_spec.rb
@@ -95,7 +95,7 @@ RSpec.describe CertificationBusinessProcess, type: :business_process do
         Strata::EventManager.publish("DeterminedNotExcluded", { case_id: certification_case.id, certification_id: certification_case.certification_id })
         certification_case.reload
 
-        # Case transitions to report_activities step is hardcoded in the business process
+        # Case transitions to external_exception_check step is hardcoded in the business process
         expect(certification_case.business_process_instance.current_step).to eq(CertificationBusinessProcess::EXTERNAL_EXCEPTION_CHECK_STEP)
         expect(certification_case.member_status).to eq(MemberStatus::AWAITING_REPORT)
         expect(certification_case).to be_open
--- a/reporting-app/spec/business_processes/certification_business_process_spec.rb
+++ b/reporting-app/spec/business_processes/certification_business_process_spec.rb
@@ -82,18 +82,21 @@ RSpec.describe CertificationBusinessProcess, type: :business_process do
         )
       end
 
-      it 'transitions to external_community_engagement_check' do
+      it 'transitions to external_exception_check' do
         # Step 1: Case has been created and is on external_exclusion_check step
         expect(certification_case.business_process_instance.current_step).to eq(CertificationBusinessProcess::EXTERNAL_EXCLUSION_CHECK_STEP)
         expect(certification_case.member_status).to eq(MemberStatus::AWAITING_REPORT)
         expect(certification_case).to be_open
+
+        # Block execution of CertificationBusinessProcess::EXTERNAL_EXCEPTION_CHECK_STEP
+        allow(ExceptionDeterminationService).to receive(:determine)
 
         # Step 2: System process determines applicant is not eligible for exclusion
         Strata::EventManager.publish("DeterminedNotExcluded", { case_id: certification_case.id, certification_id: certification_case.certification_id })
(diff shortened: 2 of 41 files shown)


reporting-app/spec/business_processes/certification_business_process_spec.rb
```diff
@@ -82,16 +82,52 @@
...
         expect(certification_case).to be_open
 
+        # Block execution of CertificationBusinessProcess::EXTERNAL_EXCEPTION_CHECK_STEP
+        allow(ExceptionDeterminationService).to receive(:determine)
+
         # Step 2: System process determines applicant is not eligible for exclusion
         Strata::EventManager.publish("DeterminedNotExcluded", { case_id: certification_case.id, certification_id: certification_case.certification_id })
         certification_case.reload
 
+        # Case transitions to report_activities step is hardcoded in the business process
```
giverm: Should this be `external_exception_check` step?

giverm (approved):
Looks good!

reporting-app/spec/models/certification_case_spec.rb
```diff
@@ -18,6 +18,33 @@
     end
   end
 
+  describe '#record_exception_determination' do
+    # The positive (excepted) path is wired end to end but currently unreachable: no exception
+    # checks are registered, so ExceptionDeterminationService never supplies reason codes. When a
+    # check story does, the case records an automated :excepted determination and closes. Reason
+    # codes are supplied by the caller (deferred to each check's story).
+    let(:reason_codes) { [ 'external_exception_met' ] }
```
jgavin-nava, line 26: Should this be added to the `VALID_REASONS`?

reporting-app/app/models/determination.rb
```diff
@@ -94,7 +94,7 @@ class Determination < Strata::Determination
   VALID_REASONS = REASON_CODE_MAPPING.values.freeze
 
   enum :decision_method, { automated: "automated", manual: "manual" }
-  enum :outcome, { compliant: "compliant", exempt: "exempt", excluded: "excluded", not_compliant: "not_compliant" }
+  enum :outcome, { compliant: "compliant", exempt: "exempt", excluded: "excluded", excepted: "excepted", not_compliant: "not_compliant" }
```
jgavin-nava, line 97: Because we add the expected outcome here we need to add a change to the member status as it would produce an invalid member status 

```
cert = FactoryBot.create(:certification)
Determination.create!(
  subject: cert,
  decision_method: "automated",
  reasons: ["age_under_19_excluded"],
  outcome: "excepted",
  determination_data: { "exception_reasons" => ["age_under_19_excluded"] },
  determined_at: Time.current
)
ms = MemberStatusService.determine(cert)
pp status: ms.status
pp valid: ms.valid?
pp errors: ms.errors.full_messages
pp dashboard: ms.dashboard_report_status
pp terminal: ms.certification_period_completed?

{status: "excepted"}
{valid: false}
{errors: ["Status is not included in the list"]}
{dashboard: "in_progress"}
{terminal: false}
```

reporting-app/spec/models/certification_case_spec.rb
jdettmannnava: This is simply a placeholder for the spec; it won't be a real reason.

jgavin-nava (approved):
LGTM

Issue #752, opened 2026-07-07
ExclusionDeterminationService: return highest-priority exclusion instead of unsorted list


giverm:
## Background
Part of the [Data Source Hierarchy spec](https://docs.google.com/document/d/1NeAEnIwRnBgtwWt_7tonPDdIZYh_KryhvQhHxXGPs6Q/edit) — the "return the best exclusion" half of the "Add Exclusion Priority Configuration" work item. Follow up to #751 (depends on the `Exclusion` priority config landing first).

Update `ExclusionDeterminationService` to evaluate exclusions in the configured priority order and return the single highest-priority match, instead of the current unsorted eligibility result.

## Acceptance Criteria
- [ ] Iterate the enabled exclusions in `Exclusion.priority_order`, stop at the first that evaluates true (stop-at-first == "best" while priority is a static list), and record/return that one.
- [ ] Record the chosen exclusion on the determination path as today; no change to the resolution/event flow beyond selecting the single best.
- [ ] Tests cover priority-ordered selection and stop-at-first behavior.

## Out-of-scope
- Data Source Registry / call-ordering (separate spec work items).

## Dependencies & Risks
- Depends on #751 (the `Exclusion` priority config).
- **Dynamic durability (PM flag, 2026-07-07):** stop-at-first equals "best" only while durability is a static priority list. When durability becomes dynamic (expected), the stop rule must be revisited — the service can no longer trust the config's static ranks and must recompute ordering per evaluation. Not a V1 blocker.

## Links & Resources
- [Data Source Hierarchy spec](https://docs.google.com/document/d/1NeAEnIwRnBgtwWt_7tonPDdIZYh_KryhvQhHxXGPs6Q/edit)

PR #758, opened 2026-07-07
Add Exclusion priority configuration (yml-configured Exclusion model)

### LINKED ISSUE #751: Add Exclusion priority configuration (yml-configured Exclusion model)
## Background
Part of the [Data Source Hierarchy spec](https://docs.google.com/document/d/1NeAEnIwRnBgtwWt_7tonPDdIZYh_KryhvQhHxXGPs6Q/edit) — the "Add Exclusion Priority Configuration" work item. CMS's HR1 interim rule requires OSCER to determine the "best" (highest-durability) exclusion for a member; this ticket adds the configuration foundation.

Create a yml-configured, deployment-overridable `Exclusion` model that returns exclusion symbols in priority order (high→low), mirroring the existing `Exemption` configuration quartet.

Exclusions are the mandatory federal minimum, so a deployment cannot disable one — the only configurable dimension is `priority` (the durability ranking). Only 3 of the 9 default exclusions (`american_indian_alaska_native`, `veteran_disability`, `pregnant`) have rules in `ExclusionRuleset` today; the config is intentionally declarative and leads the rules, with the remaining 6 wired as their rules land.

## Acceptance Criteria
- [ ] `Exclusion` utility class exposes `.all` (sorted by `:priority`), `.priority_order`, `.find`, and `.valid_values`.
- [ ] `ExclusionTypesLoader` (`extend ConfigLoading`) defines a `DEFAULTS` hash of the spec's 9 exclusions at `priority: 1–9`, a `deep_merge` override, and a `transform` requiring a distinct integer `priority` per entry (duplicate priorities raise at boot).
- [ ] `config/initializers/exclusion_types.rb` + `config/custom/exclusion_types.yml` provide the deployment override surface.
- [ ] Config shape is `{ id => { priority: N } }` — a per-entry integer field, so an override can re-rank (swap) one exclusion without restating all nine. No `enabled` flag: federal-minimum exclusions are always evaluated.
- [ ] Tests cover the defaults, override deep-merge (re-rank via swap, add), and `transform` validation failures (missing/non-integer/duplicate priority).

## Out-of-scope
- Updating `ExclusionDeterminationService` to return the highest-priority exclusion — tracked as a fast-follow (#752), foundation→implementation split.

## Dependencies & Risks
- Builds on the shared `ConfigLoading` module (#741, merged).
- Config leads the rules by design: 6 of the 9 default exclusion ids have no `ExclusionRuleset` rule yet, so they are inert until wired.

## Links & Resources
- [Data Source Hierarchy spec](https://docs.google.com/document/d/1NeAEnIwRnBgtwWt_7tonPDdIZYh_KryhvQhHxXGPs6Q/edit)

giverm:
## Ticket

Resolves #751

## Changes

- Add `Exclusion` utility class exposing `.all` (sorted ascending by `:priority`), `.priority_order` (ids high durability to low), `.valid_values`, and `.find`.
- Add `ExclusionTypesLoader`, which `extend`s the shared `ConfigLoading` module and defines: a `DEFAULTS` set of the nine specified exclusions at priorities 1-9, `merge_with_defaults` (deep-merge of the override over defaults), and `transform` (validates each entry has an integer `priority` and rejects duplicate priorities).
- Add `config/initializers/exclusion_types.rb` to load the merged, validated config into `Rails.application.config.exclusion_types` at boot.
- Add `config/custom/exclusion_types.yml`, the deployment-owned override surface (ships empty, with documented re-rank and add examples).
- Add specs for the loader (`spec/services/exclusion_types_loader_spec.rb`) and the model (`spec/models/exclusion_spec.rb`).

## Context for reviewers

The Data Source Hierarchy work needs OSCER to determine the single "best" (highest-durability) exclusion for a member. This PR is the configuration foundation for that: a yml-configured, deployment-overridable ordered list of exclusions. It deliberately mirrors the existing `Exemption` config quartet (`Exemption` + `ExemptionTypesLoader` + initializer + `config/custom` override), so if you know that pattern this will read familiar.

Two design points worth calling out:

- **Priority-only shape, no `enabled` flag.** The defaults are the CMS "specified excluded individuals" (§435.554), which are the mandatory federal minimum, so a deployment can re-rank exclusions but must not disable one. The only configurable field is an integer `priority`. Priorities must be distinct: the loader raises a boot-time `ConfigurationError` on a duplicate, because a tie would make "best exclusion" selection depend on an unstable sort.
- **The config leads the rules.** Only three of the nine defaults (`american_indian_alaska_native`, `veteran_disability`, `pregnant`) have rules in `Rules::ExclusionRuleset` today; the rest are declarative until their rules land. Note the `age_under_19` / `age_over_65` checks in `ExclusionRuleset` are intentionally absent here: per the CMS interim final rule they are mandatory exceptions (§435.553), not specified excluded individuals, so they belong to the exception work (#735/#746), not this list.

Deliberate divergences from the Exemption pattern: the integer `priority` field and priority ordering, and no `enabled` flag. It carries no i18n metadata because member-facing display of an exclusion is keyed by reason code in `MemberStatusService`, not by this config.

Out of scope: updating `ExclusionDeterminationService` to consume this ordering and return the single highest-priority exclusion is a fast-follow, #752.

## Testing

- Full suite green: 2516 examples, 0 failures; coverage 96.39% line / 82.99% branch (above the 92% / 70% gates).
- RuboCop clean on the changed files.
- The loader spec includes a boot-integration example confirming `Rails.application.config.exclusion_types` is wired by the initializer, plus validation-failure cases (non-Hash entry, missing `priority`, non-integer `priority`, duplicate `priority`) and a full load-pipeline (parse to merge to transform) example.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/services/exclusion_types_loader.rb
+++ b/reporting-app/app/services/exclusion_types_loader.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+require_relative "config_loading"
+
+# Loads exclusion-type configuration by deep-merging an optional deployment
+# override (config/custom/exclusion_types.yml, the deployment-owned override
+# surface) over the mandatory federal-minimum exclusions declared in DEFAULTS
+# below.
+#
+# Exclusions are the federal minimum set of conditions that remove a member from
+# the community-engagement requirement; a deployment cannot disable one, so
+# entries carry no `enabled` flag. The only configurable dimension is `priority`
+# (Integer, 1 = highest durability), which lets a deployment re-rank the order in
+# which exclusions are evaluated. The default ranking is OSCER-owned (updated by
+# Nava as CMS regulations evolve).
+#
+# Mirrors ExemptionTypesLoader; the YAML load/parse/error plumbing lives in the
+# shared ConfigLoading module (extend below). This loader keeps only its
+# deep-merge + transform logic, whose validation requires a distinct integer
+# `priority` per entry.
+module ExclusionTypesLoader
+  extend ConfigLoading
+
+  # Alias keeps ExclusionTypesLoader::ConfigurationError valid for existing
+  # rescues/specs; unqualified `raise ConfigurationError` in transform resolves
+  # to it via lexical scope, so no raise site changes.
+  ConfigurationError = ConfigLoading::ConfigurationError
+
+  # Default exclusion priority order (high durability -> low), per the
+  # Data Source Hierarchy spec. Only american_indian_alaska_native,
+  # veteran_disability, and pregnant have rules in Rules::ExclusionRuleset
+  # today; the rest are declarative until their rules land.
+  DEFAULTS = {
+    "american_indian_alaska_native" => { "priority" => 1 },
+    "former_foster_care" => { "priority" => 2 },
+    "veteran_disability" => { "priority" => 3 },
+    "medically_frail" => { "priority" => 4 },
+    "caretaker" => { "priority" => 5 },
+    "tanf_snap_work" => { "priority" => 6 },
+    "drug_treatment" => { "priority" => 7 },
+    "pregnant" => { "priority" => 8 },
+    "inmate" => { "priority" => 9 }
+  }.freeze
+
+  module_function
+
+  def merge_with_defaults(overrides)
+    DEFAULTS.deep_merge(overrides)
+  end
+
+  def transform(merged)
+    entries = merged.map do |id, attrs|
+      unless attrs.is_a?(Hash)
+        raise ConfigurationError, "exclusion_types.#{id}: expected Hash, got #{attrs.class}"
+      end
+      unless attrs.key?("priority")
+        raise ConfigurationError, "exclusion_types.#{id}: missing required 'priority' field"
+      end
+      unless attrs["priority"].is_a?(Integer)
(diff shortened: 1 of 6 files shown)


reporting-app/app/services/exclusion_types_loader.rb
```diff
@@ -0,0 +1,75 @@
...
+  # rescues/specs; unqualified `raise ConfigurationError` in transform resolves
+  # to it via lexical scope, so no raise site changes.
+  ConfigurationError = ConfigLoading::ConfigurationError
+
+  # Default exclusion priority order (high durability -> low), per the
+  # Data Source Hierarchy spec. Only american_indian_alaska_native,
+  # veteran_disability, and pregnant have rules in Rules::ExclusionRuleset
+  # today; the rest are declarative until their rules land.
+  DEFAULTS = {
+    "american_indian_alaska_native" => { "priority" => 1 },
```
jdettmannnava: I usually make my priorities increments of 10, so it is easier to move them around without doing an entire explicit sort. So if I wanted to move former_foster_care below caretaker, I could just make former_foster_care 55 instead of addressing all the ones in between

jdettmannnava (approved):
LGTM! I have a recommendation about priority that is not very critical

PR #761, opened 2026-07-08
ExclusionDeterminationService: return highest-priority exclusion instead of unsorted list

### LINKED ISSUE #752: ExclusionDeterminationService: return highest-priority exclusion instead of unsorted list

giverm:
## Ticket

Resolves #752

## Changes

- Select and record the single highest-priority exclusion, ranked by the configured `Exclusion.priority_order` (from #751), instead of recording every match as an unsorted list.
- Remove age from the exclusion path: drop `age_under_19` and the two age parameters from `ExclusionRuleset#eligible_for_exclusion`, plus the date-of-birth and evaluation-date extraction that only fed the age rule.
- Add `EXCLUSION_FACT_IDS`, an explicit rules-fact-name to `Exclusion` config-id map, with a fail-loud lookup and a test invariant so the naming seam cannot drift silently.
- Change `CertificationCase#record_exclusion_determination` to take reason codes directly, mirroring the existing `record_exception_determination`; the service owns selection, the model records what it is given.
- Remove `Determination.to_reason_codes` (it produced the old unsorted list).

## Context for reviewers

The background exclusion check previously recorded every exclusion a member matched. The Data Source Hierarchy spec calls for OSCER to record the single "best" (most durable) exclusion, so this returns the one with the lowest priority number in the #751 config. Selection is a plain `min` over `Exclusion.priority_order`, so a deployment can re-rank exclusions and change the winner with no code change (a spec covers this via a config override).

Age is dropped intentionally, confirmed with Jeff: the state does not send members outside the community-engagement age range, and age is a CMS 435.553 mandatory exception rather than an exclusion. The age reason codes are deliberately left in `REASON_CODE_MAPPING` and the locales (they still serve the display and exemption paths); only the exclusion rule and its inputs are removed.

The `EXCLUSION_FACT_IDS` bridge exists because the rules-engine fact names and the `Exclusion` config ids are different vocabularies (`is_veteran_with_disability` vs `veteran_disability`, and the fact's "or" that the AIAN id drops), so they are mapped explicitly rather than string-derived; `exclusion_priority` raises naming the fact and id if the seam ever drifts.

Forward context, within the Data Source Hierarchy work this is part of: a later story adds mandatory exceptions, where a condition that applied during a lookback month but no longer applies is credited as met for that month. OSCER makes that time-based judgment, and those exception results rank below all exclusions on the same priority ladder. The single-best selection added here is the mechanism that ladder extends: the exception band and the configurable ordering of external data-source calls build on it rather than replace it.

## Testing

- Full suite: 2552 examples, 0 failures. Coverage 96.45% line / 83.46% branch. RuboCop clean.
- New service specs cover priority-ordered selection, stop-at-first (recorded `reasons` is a single-element array), a config override proving ordering is config-driven, and the fail-loud drift guard.
- Added `EXCLUSION_FACT_IDS` invariant specs asserting every bridged id is configured with a priority and every key maps to a reason code.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/determination.rb
+++ b/reporting-app/app/models/determination.rb
@@ -118,11 +118,6 @@ class Determination < Strata::Determination
       .order("subject_id, created_at DESC")
   }
 
-  def self.to_reason_codes(eligibility_fact)
-    eligibility_fact_reasons = eligibility_fact.reasons.select { |reason| reason.value }.map(&:name).map(&:to_sym)
-    eligibility_fact_reasons.map { |reason| REASON_CODE_MAPPING[reason] }
-  end
-
   # CE automated/manual payload uses +determination_data["calculation_type"]+ with string keys from JSON.
   # @return [String, nil] e.g. +CALCULATION_TYPE_INCOME_BASED+, +CALCULATION_TYPE_HOURS_BASED+, +CALCULATION_TYPE_EXTERNAL_CE_COMBINED+
   def ce_calculation_type
--- a/reporting-app/app/models/determination.rb
+++ b/reporting-app/app/models/determination.rb
@@ -89,8 +89,6 @@ class Determination < Strata::Determination
   ).freeze
 
   EXEMPTION_REASONS = REASON_CODE_MAPPING.values_at(
-    :age_under_19,
-    :age_over_65,
     :is_pregnant,
     :is_american_indian_or_alaska_native,
     :exemption_request_compliant,
--- a/reporting-app/app/models/rules/exclusion_ruleset.rb
+++ b/reporting-app/app/models/rules/exclusion_ruleset.rb
@@ -5,15 +5,6 @@ module Rules
   class ExclusionRuleset < Strata::Rules::MedicaidRuleset
     AMERICAN_INDIAN_OR_ALASKA_NATIVE = [ "american_indian_or_alaska_native", "american_indian", "alaska_native" ].freeze
 
-    # Rules-engine fact name => Exclusion config id (its priority source). The two
-    # use different vocabularies (is_veteran_with_disability vs veteran_disability;
-    # the fact's "or" the AIAN id drops), so map explicitly — don't string-derive.
-    EXCLUSION_FACT_IDS = {
-      is_american_indian_or_alaska_native: :american_indian_alaska_native,
-      is_veteran_with_disability: :veteran_disability,
-      is_pregnant: :pregnant
-    }.freeze
-
     def is_pregnant(pregnancy_status)
       return if pregnancy_status.nil?
 
--- a/reporting-app/app/models/rules/exclusion_ruleset.rb
+++ b/reporting-app/app/models/rules/exclusion_ruleset.rb
@@ -1,16 +1,18 @@
 # frozen_string_literal: true
 
 module Rules
-  # Implements eligibility rules for age-based Medicaid exclusions.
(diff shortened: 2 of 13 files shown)


reporting-app/app/models/rules/exclusion_ruleset.rb
```diff
@@ -1,16 +1,18 @@
...
+  # Eligibility rules for the community-engagement exclusions.
   class ExclusionRuleset < Strata::Rules::MedicaidRuleset
     AMERICAN_INDIAN_OR_ALASKA_NATIVE = [ "american_indian_or_alaska_native", "american_indian", "alaska_native" ].freeze
 
-    def age_under_19(age)
-      return if age.nil?
-
-      age < 19
-    end
+    # Rules-engine fact name => Exclusion config id (its priority source). The two
```
jdettmannnava: Given that these fact ids are also mapped in Determination::REASON_CODES, I think it would be better to update the Exclusion config to keep everything in sync?

reporting-app/app/models/determination.rb
jdettmannnava: I am not sure I understand why these remain?

jdettmannnava (approved):
I have a couple questions, so only change if you don't have good answers!

reporting-app/app/models/rules/exclusion_ruleset.rb
giverm: Moved to the config, per your suggestion. Let me know if this matching what you were thinking

reporting-app/app/models/determination.rb
giverm: Dropped the age exemption reasons

reporting-app/app/models/rules/exclusion_ruleset.rb
jdettmannnava: Oh, I was just thinking of renaming the config keys from "pregnant" to "is_pregnant", etc

PR #765, opened 2026-07-09
Check exceptions on months instead of boolean in member data

### LINKED ISSUE #764: Optional Exception determination based on date
## Background
We need to use date matching for optional exceptions, as the application of exceptions is based on lookback period, which api users will not be able to match.

## User Story
As a user, I want optional exception mapping to be date-based, so that it matches likely data.

## Acceptance Criteria

- [ ] Optional exceptions apply when they exist within the lookback period
- [ ] Optional exceptions do not apply when they exist outside the lookback period

## Out-of-scope (optional)  
Mandatory exceptions are out of scope.

jdettmannnava:
## Ticket

Resolves #764 

## Changes

- Member Data takes list of dates for exceptions rather than boolean
- ExceptionDeterminationService checks months in MemberData against certification months in Requirements to determine if exception applies.
- Demo updated to set member data appropriately.

## Context for reviewers

Realized we need to validate exceptions against the dates in which the events took place instead of just generally.

## Testing
Updated specs.

Manually tested using demo:
- No exception had no affect
- Exception set member data appropriately and led to proper case determination

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/services/exception_determination_service.rb
+++ b/reporting-app/app/services/exception_determination_service.rb
@@ -8,7 +8,7 @@
 class ExceptionDeterminationService
   include Strata::VirtualActor
 
-  # Each symbol names a private check method taking certification and returning its reason code when
+  # Each symbol names a private check method taking member_data and lookback months and returning its reason code when
   # the exception applies (and is enabled), or nil otherwise. Add a check by adding its symbol here
   # and defining the matching method. Order is the evaluation order; the first applicable check wins.
   EXCEPTION_CHECKS = %i[
--- a/reporting-app/app/services/exception_determination_service.rb
+++ b/reporting-app/app/services/exception_determination_service.rb
@@ -8,7 +8,7 @@
 class ExceptionDeterminationService
   include Strata::VirtualActor
 
-  # Each symbol names a private check method taking member_data and returning its reason code when
+  # Each symbol names a private check method taking certification and returning its reason code when
   # the exception applies (and is enabled), or nil otherwise. Add a check by adding its symbol here
   # and defining the matching method. Order is the evaluation order; the first applicable check wins.
   EXCEPTION_CHECKS = %i[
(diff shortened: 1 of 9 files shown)


reporting-app/spec/services/exception_determination_service_spec.rb
```diff
@@ -89,50 +115,27 @@
...
     context 'when more than one exception check would apply' do
-      let(:certification) do
-        create(:certification, member_data: build(
-          :certification_member_data,
-          receiving_inpatient_medical_care: true,
-          resides_in_declared_emergency_county: true
-        ))
-      end
+      let(:dates_receiving_inpatient_medical_care) { [ cert_date - 3.months ] }
+      let(:dates_traveling_for_medical_care) { [ cert_date - 3.months ] }
```
giverm, line 139: Would it be worth adding some test dates that fall somewhere mid month, rather than only on the first of the month?

reporting-app/app/services/exception_determination_service.rb
```diff
@@ -48,42 +48,57 @@ def applicable_exception_reason_codes(certification)
...
     end
 
-    # @return [String, nil] the reason code when the member is receiving inpatient medical care and
+    # @return [String, nil] the reason code when the member was receiving inpatient medical care and
     #   the exception is enabled, otherwise nil.
-    def inpatient_medical_care(member_data)
+    def inpatient_medical_care(member_data, certifiable_months)
       return unless ExternalException.enabled?(:inpatient_medical_care)
-      return unless member_data.receiving_inpatient_medical_care
+      return unless member_data.dates_receiving_inpatient_medical_care.present?
```
giverm, line 62: If `member_data.dates_receiving_inpatient_medical_care` receives an invalid date string (`["not-a-date"]` the value becomes `[nil]` and `[nil].present?` returns `true`. Then `beginning_of_month` will raise a no method error.

giverm (approved):

reporting-app/app/services/exception_determination_service.rb
jdettmannnava: Added `compact` before mapping

Issue #766, opened 2026-07-10
Represent lapsing exclusion signals as lookback dates in MemberData


giverm:
Part of the exclusion/exception determination work under epic #732 (Data Source Hierarchy). Prerequisite for #767 and #768. Depends on #761.

## Background

`ExclusionDeterminationService` evaluates exclusion signals as point-in-time booleans on `Certifications::MemberData`. Some exclusion conditions, such as incarceration, apply only for a specific date range, and whether the member is excluded depends on when those dates fall:

- If the condition is active during the certification month, the member is excluded.
- If the condition is not active during the certification month but its dates fall within the certification lookback, it counts as a mandatory exception rather than an exclusion.

A boolean cannot express this distinction. This introduces date-based representation for those signals, following the pattern used for external exceptions in #765, which stores the signals as arrays of dates on `MemberData` and matches them at month granularity.

## User Story

As a Medicaid program processing a certification, I want exclusion signals stored as dates, so the determination can tell whether a condition is active during the certification month (an exclusion) or is no longer active but fell within the lookback (a mandatory exception).

## Acceptance Criteria

- [ ] `Certifications::MemberData` represents a lapsing exclusion signal (incarceration) as an array of dates, following the `dates_*` pattern from #765.
- [ ] `ExclusionDeterminationService` determines the member excluded only when the signal includes the certification month (the month of `certification_requirements.certification_date`), matching the month-level comparison used for exceptions.
- [ ] A signal that does not include the certification month does not produce an exclusion, even when its dates fall within the lookback.
- [ ] Demo tooling can seed the date signal, as in #765.

## Out-of-scope

- Emitting the mandatory exception for a signal that is not active during the certification month but whose dates fall within the lookback. That belongs to the ordered determination flow tracked in #767.
- Sourcing incarceration dates from an external data source, handled under the data-source contract work (#755). This issue populates the signal from `MemberData`.

## Dependencies & Risks

- Depends on #761 (return the highest-priority exclusion). `main`'s `ExclusionDeterminationService` still evaluates the pre-#761 aggregate fact, so this builds on the post-#761 shape.
- Builds on the pattern merged in #765 and #757.

## Links & Resources

- Data Source Hierarchy tech spec, §"Exceptions from former exclusions" (linked from epic #732)
- Related: #765, #757 (exception date pattern), #761 (highest-priority exclusion)

PR #769, opened 2026-07-10
OSCER-766: base pregnancy exclusion on (real or expected) parturition date

jdettmannnava:
## Ticket

First pass at #766

## Changes

Converts the pregnancy signal used by `ExclusionDeterminationService` from a boolean to a date, bringing exclusion member data in line with the date-based approach already used for exception member data.

- **`Certifications::MemberData`** — replaced `pregnancy_status` (boolean) with `pregnancy_due_or_parturition_date` (date): the expected due date if the member is currently expecting, or the actual birth date if in the past.
- **`Rules::ExclusionRuleset#is_pregnant`** — now takes `(pregnancy_due_or_parturition_date, certification_date)` and applies a 12-month exclusion window (`POSTPARTUM_EXCLUSION_MONTHS`) evaluated against the certification date.
- **`ExclusionDeterminationService`** — sets the due/parturition date and the certification date (from `certification_requirements.certification_date`) as facts for the rules engine.
- **Demo create form** — the `pregnancy_status` checkbox now writes the certification date into `pregnancy_due_or_parturition_date`.
- **Batch upload (`UnifiedRecordProcessor`)** — a truthy `pregnancy_status` CSV value records the certification date as the parturition date. The check is case-insensitive and accepts `yes`/`true`. 
Only the pregnancy signal changed — AIAN and veteran-disability exclusions and the priority-ordering mechanism are untouched. The demo, API, and batch upload input fields remain `pregnancy_status`; no documentation was modified.

## Context for reviewers

A member is pregnancy-excluded when the certification date is on or before the end of the 12-month window (`pregnancy_due_or_parturition_date + 12.months`).

The window end is evaluated at **month granularity**: as long as the window reaches the first day of the certification date's month, the exclusion covers that entire month. So a certification date of June 20 is still excluded when the window ends June 5, but a window ending in May does not carry into June.

The input surfaces intentionally stay a `pregnancy_status` boolean (checkbox / CSV column) and translate to the new date field by writing the certification date — this keeps the member/staff-facing inputs unchanged while the exclusion evaluates a date.

For the batch upload, the documentation says use "yes/no", but the sample doc uses "FALSE" for pregnancy status, so I went with as inclusive a definition as possible. This also fixes a latent bug: the previous boolean cast treated `no` as truthy, since `no` is not in ActiveModel's false-values set.

## Testing

Full suite green (2588 examples, 0 failures), `make lint` reports no offenses.

- **`Rules::ExclusionRuleset`** — rewrote `#is_pregnant` coverage: nil guards, currently-expecting, postpartum, exact-12-month boundary, month-granularity (window ends earlier in the same month → excluded; window ends the prior month → not excluded), and lapsed.
- **`ExclusionDeterminationService`** — pregnancy cases updated to the date field, with future-due and postpartum excluded cases plus a lapsed (>12 months) not-excluded case.
- **`UnifiedRecordProcessor`** — new cases confirming `yes`/`YES`/` Yes `/`true`/`TRUE` record the certification date and `no`/`false`/`FALSE`/`""` leave it unset.
- Demo request spec and member-data factory updated for the renamed field.

```
make test   # 2588 examples, 0 failures
make lint   # 541 files inspected, no offenses detected
```
### Manually tested demo
- Pregnancy flag added appropriate member data, case exempt
- Updated pregancy flag for 15 months prior to certification date; had appropriate member data; was not exempt

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/forms/demo/certifications/create_form.rb
+++ b/reporting-app/app/forms/demo/certifications/create_form.rb
@@ -69,7 +69,7 @@ module Demo
         member_data = ::Certifications::MemberData.new(member_data).tap do |md|
           md.name = member_name if member_name.present?
           md.date_of_birth = date_of_birth if date_of_birth.present?
-          md.pregnancy_status = pregnancy_status if pregnancy_status.present?
+          md.pregnancy_due_or_parturition_date = certification_date if pregnancy_status
           md.race_ethnicity = race_ethnicity if race_ethnicity.present?
           md.va_icn = va_icn if va_icn.present?
           apply_external_exception(md, certification_requirements.months_that_can_be_certified)
--- a/reporting-app/app/models/certifications/member_data.rb
+++ b/reporting-app/app/models/certifications/member_data.rb
@@ -70,11 +70,13 @@ class Certifications::MemberData < ValueObject
   attribute :contact, ContactData.to_type
   attribute :name, ActiveModel::Type::Json.new(Strata::Name)
   attribute :date_of_birth, :date
-  attribute :race_ethnicity, :string
 
   attribute :payroll_accounts, :array, of: PayrollAccount.to_type
   attribute :activities, :array, of: Activity.to_type
-  attribute :pregnancy_status, :boolean, default: false
+
+  # Exclusion signals evaluated by ExclusionDeterminationService
+  attribute :pregnancy_due_or_parturition_date, :date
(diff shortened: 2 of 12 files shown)


jgavin-nava (approved):
I think we need to update the openAPI docs as it still talks about the `pregnancy_status` on there

PR #783, opened 2026-07-14
Jdettmann/oscer 766 additional exclusions

### LINKED ISSUE #766: Represent lapsing exclusion signals as lookback dates in MemberData

jdettmannnava:
## Ticket

Resolves #766

## Changes

- Added checks for missing exclusions
- Replaced automatic veteran disability check with boolean in member data
- Add checkboxes to demo form for new exclusions
- Updated mistaken fix from last PR in openapi doc
- Block SimpleCov under test-watch

## Context for reviewers

To finish out the date testing, all exclusions were added to the data determination service. The veteran disability check based on icn was replaced with a member_data check, because the check by icn will be a data source check.

Also, updating the api doc for pregnancy was wrong, as the api still takes pregnancy status, it just updates the member data differently.

## Testing

Automated tests pass.

Manual checks for each new exclusion verified. Verified order of checks worked by checking two.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/services/exclusion_determination_service.rb
+++ b/reporting-app/app/services/exclusion_determination_service.rb
@@ -30,18 +30,18 @@ class ExclusionDeterminationService
       ruleset = Rules::ExclusionRuleset.new
       engine = Strata::RulesEngine.new(ruleset)
 
-      pregnancy_due_or_parturition_date = extract_pregnancy_due_or_parturition_date(certification)
+      pregnancy_due_or_parturition_date = extract_attribute(certification, :pregnancy_due_or_parturition_date)
       certification_date = certification.certification_requirements.certification_date
-      race_ethnicity = extract_race_ethnicity(certification)
-      veteran_with_disability = extract_veteran_with_disability(certification)
-      was_in_foster_care = extract_was_in_foster_care(certification)
-      date_of_birth = extract_date_of_birth(certification)
-      currently_medically_frail = extract_currently_medically_frail(certification)
-      dates_caretaking_infirm = extract_dates_caretaking_infirm(certification)
-      dependent_children_birth_dates = extract_dependent_children_birth_dates(certification)
-      meeting_tanf_or_snap_work = extract_meeting_tanf_or_snap_work(certification)
-      dates_in_drug_treatment = extract_dates_in_drug_treatment(certification)
-      dates_incarcerated = extract_dates_incarcerated(certification)
+      race_ethnicity = extract_attribute(certification, :race_ethnicity)
+      veteran_with_disability = extract_attribute(certification, :veteran_with_disability)
+      was_in_foster_care = extract_attribute(certification, :was_in_foster_care)
+      date_of_birth = extract_attribute(certification, :date_of_birth)
+      currently_medically_frail = extract_attribute(certification, :currently_medically_frail)
+      dates_caretaking_infirm = extract_attribute(certification, :dates_caretaking_infirm)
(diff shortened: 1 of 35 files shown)


reporting-app/app/services/exclusion_determination_service.rb
```diff
@@ -64,22 +67,10 @@ def exclusion_priority(fact_name)
...
+    def extract_attribute(certification, attribute)
       return nil unless certification.member_data
 
-      certification.member_data.race_ethnicity
-    end
-
-    def extract_veteran_disability_status(certification)
-      return nil unless certification.member_data&.va_icn.present?
-
-      VeteranDisabilityService.new.get_disability_rating(icn: certification.member_data.va_icn)
```
jgavin-nava, line 82: Just so I am understanding this correctly we are removing this because we would have already ran the checks to external service and then it just checks the member data since it has already been done and doesnt need to be called again?

jgavin-nava (approved):
LGTM jsut a clarifying question for my understanding

reporting-app/app/services/exclusion_determination_service.rb
jdettmannnava: We are removing this here because we are only going to run this if no other exclusion with a higher priority is in the member data. I was planning on adding this back in with the DataSource configuration.

Issue #754, opened 2026-07-07
Verification data source adapter contract + result object


jgavin-nava:
**Ticket A** in the verification data-source work under epic #732 (Data Source Hierarchy). Blocks B (VA refactor, #755). Sibling: Data Source registry (#756) — distinct from Exclusion priority config (#751 / #758).

## Background

OSCER is building a verification hierarchy that calls external data sources (VA disability today; more later) to derive exclusion / exception / community-engagement outcomes for a member certification. Today there is no shared contract: `VeteranDisabilityService#get_disability_rating(icn:)` returns raw VA JSON or `nil` (fail-open), and interpretation of that JSON lives outside the service (in `Rules::ExclusionRuleset#is_veteran_with_disability`). We want every data source to conform to one adapter interface so an orchestrator (built later, out of scope here) can call them uniformly and audit them consistently.

This ticket defines the interface and its result object only — no orchestrator, no rules-engine changes, no `Determination` mapping, and **no priority/ordering logic**.

## Config axes (do not conflate)

| Config | Owns | Ticket |
|--------|------|--------|
| **Exclusion** | Durability ranking of exclusion *outcomes* | #751 / #758 (`Exclusion`, `ExclusionTypesLoader`) |
| **Exception** | Enable/disable optional exceptions | Exception config work |
| **Data Source** | Which *sources* exist, which exclusion/exception/CE checks each can make, call order for exceptions + CE | **#756** (reopened; was incorrectly closed as a dup of #751) |

#754 is the **adapter code contract**. #756 is the **YAML registry of sources**. Adapters implement this contract; the registry names which adapters are in play and what checks they may declare.

## User Story

As an OSCER engineer adding a new external verification data source, I want a single adapter contract and result object so that each source returns a predictable shape (outcome symbols + audit data), handles its own preconditions and errors, and can be tested against shared expectations.

## Shared outcome vocabulary (aligns with #758 on `main`)

Outcome symbols an adapter emits are **not** free-form and are **not** ruleset method names or `Determination` reason codes. For the exclusion category, symbols **MUST be `Exclusion` ids** — the canonical ids from #758:

`american_indian_alaska_native`, `former_foster_care`, `veteran_disability`, `medically_frail`, `caretaker`, `tanf_snap_work`, `drug_treatment`, `pregnant`, `inmate`.

[...]

PR #784, opened 2026-07-14
Add verification Data Source registry (OSCER-756)

### LINKED ISSUE #756: Data Source registry (YAML config for verification adapters)
**Ticket C** in the verification data-source work under epic #732 (Data Source Hierarchy). Distinct from Exclusion priority config (#751 / #758) and Exception enable/disable config. Depends on the adapter contract (#754) for vocabulary alignment; orchestrator consumption is a later slice.

## Background

Three config axes exist in the Data Source Hierarchy work (as clarified by @jdettmannnava / @giverm on #754):

| Config | Purpose | Example |
|--------|---------|---------|
| **Exclusion** (#751 / #758) | Rank exclusion *outcomes* by durability (priority). No `enabled` — federal minimum, always evaluated. | `Exclusion.priority_order` |
| **Exception** | Enable/disable *optional* exceptions. | Exception enable flag(s) |
| **Data Source** (this ticket) | Register *sources*: which exclusion / exception / CE checks each can make, and **call order for exceptions and CE**. | Per-source adapter + declared checks + order |

This ticket is the **Data Source** axis. It was previously closed as a duplicate of #751 — that was incorrect. Exclusion config ranks outcomes; this config registers where to look for data and in what order to call sources for exception/CE paths. Exclusion evaluation order stays owned by `Exclusion` durability (#758 / #752), not by this registry.

Reuse OSCER's existing config pattern (`ConfigLoading` + `DEFAULTS` + `config/custom/*.yml` deep-merge + boot `ConfigurationError`), mirroring `ExemptionTypesLoader` / `ExclusionTypesLoader`. Per-deployment `config/custom/` only (no multi-state profiles).

## User Story

As a deployment maintainer, I want to configure which external data sources exist, which exclusion/exception/CE checks each can make, and (for exceptions and CE) the order those sources are called — via YAML over OSCER defaults — so sources can be added/tuned without code changes and invalid config fails loudly at boot.

## Acceptance Criteria

- [ ] `VerificationDataSourcesLoader` (name flexible) mirrors `ExclusionTypesLoader` / `ExemptionTypesLoader`: OSCER-owned `DEFAULTS` + optional `config/custom/verification_data_sources.yml`, deep-merged with override keys winning; reuses `ConfigLoading`.
- [ ] Each source entry declares at least: identity/key, `adapter_class` (constantizes at boot), and which **checks by category** it can make — exclusion ids and/or exception ids and/or CE check ids — using the shared vocabulary (`Exclusion.valid_values` for exclusion; Exception registry when it lands; CE ids as defined by the hierarchy work).
- [ ] Config supports **call ordering for exception and CE sources** (and documents that **exclusion** call/selection order is *not* configured here — it is durability-driven via `Exclusion.priority_order`).
- [ ] Boot validation raises `ConfigurationError` naming the offending entry when: malformed YAML / missing required keys; `adapter_class` does not constantize; a declared check id is not in the corresponding category registry; duplicate/conflicting source entries; invalid order fields (as defined by the config shape).
- [ ] Initializer sets `Rails.application.config.verification_data_sources` (mirroring `exclusion_types` / `exemption_types` initializers).
- [ ] `config/custom/verification_data_sources.yml` ships empty (`{}`) with a documented header (examples for add / re-order / declare checks).
- [ ] Specs mirror `exclusion_types_loader_spec.rb` / `exemption_types_loader_spec.rb` edge cases, plus registry-subset and adapter_class loadability.

## Out-of-scope

- Implementing adapters themselves (#754 contract, #755 VA reference).
- Orchestrator wiring that *consumes* this registry to call sources (skip-when-met, audit-always, stop rules) — separate under #732.
- Changing Exclusion durability ranking (#751/#758) or #752 stop-at-first in `ExclusionDeterminationService`.
- Multi-state-profile config shape.

## Dependencies & Risks

- **Vocabulary:** exclusion check ids ⊆ `Exclusion.valid_values` (#758). Exception/CE ids need a settled registry before those categories can be validated the same way.
- **Depends on #754** for the adapter contract adapters named in `adapter_class` must implement.
- **Credentials drift:** `enabled` / present sources vs missing `VA_*` env — loader should not crash; runtime skip/error belongs to the adapter.

## Open questions

- Exact YAML shape for “checks by category” vs a flat outcomes list — pin to the Data Source Hierarchy tech spec if it already defines a schema.
- Whether `enabled` is per-source (Exception-like) or always-on for federally required sources.

## Links & Resources

- Clarification thread: #754 (comments from @giverm / @jdettmannnava)
- Contrast (not this ticket): `reporting-app/app/models/exclusion.rb`, `reporting-app/app/services/exclusion_types_loader.rb`, #751 / #758
- Patterns to mirror: `reporting-app/app/services/config_loading.rb`, `ExemptionTypesLoader`, `ExclusionTypesLoader`
- Epic: #732
- Sibling: #754 (contract), #755 (VA adapter)

jgavin-nava:
## Ticket

Resolves [OSCER-756](https://github.com/navapbc/oscer/issues/756)

## Changes

- Add `VerificationDataSourcesLoader` with OSCER-owned `DEFAULTS`, deep-merge over optional `config/custom/verification_data_sources.yml`, and shared `ConfigLoading` YAML plumbing (mirrors exclusion/exemption loaders).
- Register the shipped VA disability-rating source (`Verification::Adapters::VaDisabilityRating`). Config entries are thin: `enabled`, `adapter_class`, `exception_order`, `ce_order`.
- Outcomes are owned by the adapter via `.declared_outcomes` (not YAML `checks`). Boot validation requires those symbols to exist in `Exclusion` / `ExternalException` registries; enabled sources must declare at least one outcome.
- Reject YAML `checks` and `exclusion_order` keys with clear `ConfigurationError`s; reject `exception_order` / `ce_order` when the adapter emits no outcome in that category.
- Boot wiring via `config/initializers/verification_data_sources.rb`: structural `transform` in the initializer body; `validate_registry!` in `to_prepare`.
- Ship empty override file (`{}`) with deployment docs (consumers must sort by `exception_order` / `ce_order`).
- Specs for merge/transform/registry validation, full pipeline, and initializer wiring.

## Context for reviewers

- Addresses review feedback: YAML owns enablement/wiring/order; adapter classes declare response codes; category mapping stays internal via existing registries (flat `.declared_outcomes`, not per-category YAML).
- **AC note:** the issue originally described YAML `checks` by category. After review we intentionally moved outcomes onto `.declared_outcomes` so config does not re-declare what the class already owns.
- **Out of scope (per ticket):** orchestrator consumption, adapter implementation (#754/#755 already landed), Exclusion durability ranking.
- `enabled` and `Rails.application.config.verification_data_sources` are **not** consumed by determination services yet — toggling `enabled: false` is a no-op at runtime until the orchestrator slice.
- CE membership for declared outcomes is not separate yet (no CE registry); unknown outcomes fail boot rather than being treated as provisional CE. Non-nil `ce_order` therefore fails boot until a CE registry exists.
- Exception vocabulary uses `ExternalException.all` (including disabled); evaluation gating remains `ExternalException.enabled?`.

## Testing

```bash
cd reporting-app
make lint
make test args='spec/services/verification_data_sources_loader_spec.rb'
```

- Confirm boot Spec sees `:va_disability_rating` with `adapter_class` pointing at `VaDisabilityRating` and no `:checks` key.
- Confirm `VaDisabilityRating.declared_outcomes` is `[ :is_veteran_with_disability ]`.
- Confirm `config/custom/verification_data_sources.yml` remains `{}` with no live overrides.

♻️ Environment destroyed ♻️


reporting-app/app/services/verification_data_sources_loader.rb
```diff
@@ -0,0 +1,247 @@
...
+  # each have a backing registry; :ce does not yet (its ids are shape-checked but
+  # not membership-checked until a CE registry lands).
+  CATEGORIES = %i[exclusion exception ce].freeze
+
+  # OSCER-owned defaults. Deployments customize via the override file, not by
+  # editing this constant. The VA disability-rating source lands here as the
+  # first real registered source (OSCER-755 adapter, OSCER-756 registry).
+  DEFAULTS = {
+    "va_disability_rating" => {
+      "enabled" => true,
```
jdettmannnava, line 50: I think this should be simplified to "enable" and "adapter_class". The verification should make sure the that the class supplies the checks.

jdettmannnava (commented):
On second thoughts, I think keeping the deep-merge structure is fine.  I also think the "enabled" flag could be useful, especially as it could be toggled with environment variables if we want.
But the classes themselves should declare the response codes. I am unsure if they need to be split among the three types (exclusion, exception, ce), or if we should just let our internal mapping figure such things out.

reporting-app/app/services/verification_data_sources_loader.rb
jgavin-nava: Agreed — implemented on this branch:

- YAML keeps deep-merge + `enabled` / `adapter_class` / `exception_order` / `ce_order`
- Outcomes live on the adapter via `.declared_outcomes` (YAML `checks` is rejected)
- Boot maps flat outcomes onto Exclusion / ExternalException internally (no per-category split on the class for now)
- Enabled sources must declare at least one outcome; order fields must match a category the adapter actually emits

reporting-app/app/services/verification_data_sources_loader.rb
```diff
@@ -0,0 +1,244 @@
...
+  # OSCER-owned defaults. Deployments customize via the override file, not by
+  # editing this constant. The VA disability-rating source lands here as the
+  # first real registered source (OSCER-755 adapter, OSCER-756 registry).
+  # Outcomes for that adapter live on
+  # Verification::Adapters::VaDisabilityRating.declared_outcomes — not here.
+  DEFAULTS = {
+    "va_disability_rating" => {
+      "enabled" => true,
+      "adapter_class" => "Verification::Adapters::VaDisabilityRating",
+      "exception_order" => nil,
```
jdettmannnava: I don't think this is going to work, as both exceptions and community-engagement data sources will be checked in the same business process and may alternate.

reporting-app/app/services/verification_data_sources_loader.rb
```diff
@@ -0,0 +1,232 @@
...
+# owns which outcome symbols it may emit via {.declared_outcomes}. Boot
+# validation confirms those outcomes exist in the Exclusion / ExternalException
+# registries (CE membership lands with a CE registry later).
+#
+# Call order among verification sources is configured here via the per-source
+# order integer. Exception and community-engagement sources may interleave in
+# the same business process, so there is one source call order rather than
+# separate category-specific orders. Exclusion call/selection order is NOT
+# configured here: Exclusion.priority_order (config/custom/exclusion_types.yml)
+# owns it, so an exclusion_order key on an entry is rejected at boot.
```
jdettmannnava: Unsure why we are worrying about keys that no longer exist?
Should document setting order keys should only be done for data sources that do not return exclusions.

reporting-app/app/services/verification_data_sources_loader.rb
```diff
@@ -0,0 +1,232 @@
...
+    end
+  end
+
+  # ---- structural (boot-body-safe) ------------------------------------------
+
+  def transform_entry(id, attrs)
+    unless attrs.is_a?(Hash)
+      raise ConfigurationError, "verification_data_sources.#{id}: expected Hash, got #{attrs.class}"
+    end
+    reject_exclusion_order!(id, attrs)
```
jdettmannnava: Unsure why we are checking for keys that aren't used anywhere?

reporting-app/app/services/verification_data_sources_loader.rb
```diff
@@ -0,0 +1,232 @@
...
+    return nil unless attrs.key?(field)
+    value = attrs[field]
+    return nil if value.nil?
+    unless value.is_a?(Integer)
+      raise ConfigurationError, "verification_data_sources.#{id}: '#{field}' must be an Integer or null, got #{value.class}"
+    end
+    value
+  end
+
+  # Exclusion call/selection order is owned by Exclusion.priority_order, not this
```
jdettmannnava: I don't see the point of these reject methods?

jgavin-nava: Agreed something cursor likes to do that I looked over this time is to fail safely when we remove something since it is thinking there could be something using the old way but we dont have to worry about it. Thanks for catching the unnecessary code

jdettmannnava (approved):
LGTM

PR #786, opened 2026-07-15
OSCER-785: Add mandatory exceptions from exclusions

### LINKED ISSUE #785: Add Mandatory exceptions to ExceptionDeterminationService
## Background
The ExceptionDeterminationService only acts on optional exceptions. Mandatory exceptions based on former exclusions need to be added.

## User Story

As an applicant, I want all exceptions checked before demanding I fill out a form, so that my administrative burden is not so bad.

## Acceptance Criteria

- [ ] All exclusions that were valid in lookback period are marked as exceptions by the ExceptionDeterminationService

## Out-of-scope (optional)  
- Handling such exceptions in the UI

jdettmannnava:
## Ticket

Resolves #785

## Changes

- `ExceptionDeterminationService`: added 5 mandatory exception checks migrated from the exclusion ruleset (pregnancy, former foster care, caretaker, drug treatment, inmate), and reordered the checks so all mandatory checks run before the optional ones.
- `Determination`: added the 5 corresponding `*_excepted` reason codes.
- Added spec coverage for the new checks and updated the `VALID_REASONS` guard.

## Context for reviewers

A member is excepted whenever the corresponding exclusion would have been valid in any certifiable month. For checks that are monotonic in time — an age or window that only closes as months pass — it is enough to test the earliest certifiable month, as `age_under_19` already does. `inmate` is the one exception: its window is bounded on both ends, so every certifiable month is tested against every incarceration window. Each check reuses the `Rules::ExclusionRuleset` constants so the exclusion and exception sides stay in sync.

Four exclusion checks were deliberately not migrated: American Indian / Alaska Native, veteran with disability, medically frail, and TANF/SNAP work. This may change based on API changes.

## Testing

`make lint test` passes (2869 examples, 0 failures; line coverage 96.56%, branch coverage 83.84%).

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/determination.rb
+++ b/reporting-app/app/models/determination.rb
@@ -82,6 +82,14 @@ class Determination < Strata::Determination
     denial_response_not_convincing: "denial_response_not_convincing",
     # External-exception reason codes (see ExceptionDeterminationService). "Excepted" is a
     # distinct outcome from "excluded"/"exempt" — do not conflate the three.
+    #
+    # Mandatory exceptions migrated from the exclusion ruleset carry the exclusion's reason code with
+    # "excluded" replaced by "excepted" (e.g. pregnancy_excluded -> pregnancy_excepted).
+    was_pregnant: "pregnancy_excepted",
+    former_foster_care_excepted: "former_foster_care_excepted",
+    was_caretaker: "caretaker_excepted",
+    was_in_drug_treatment: "drug_treatment_excepted",
+    was_inmate: "inmate_excepted",
     age_was_under_19: "age_under_19_excepted",
     receiving_inpatient_medical_care: "inpatient_medical_care_excepted",
     resides_in_declared_emergency_county: "declared_emergency_county_excepted",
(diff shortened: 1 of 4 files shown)


reporting-app/app/models/determination.rb
```diff
@@ -82,6 +82,14 @@ class Determination < Strata::Determination
     denial_response_not_convincing: "denial_response_not_convincing",
     # External-exception reason codes (see ExceptionDeterminationService). "Excepted" is a
     # distinct outcome from "excluded"/"exempt" — do not conflate the three.
+    #
+    # Mandatory exceptions migrated from the exclusion ruleset carry the exclusion's reason code with
+    # "excluded" replaced by "excepted" (e.g. pregnancy_excluded -> pregnancy_excepted).
+    was_pregnant: "pregnancy_excepted",
+    former_foster_care_excepted: "former_foster_care_excepted",
```
jgavin-nava: Should this match the other ones?

```suggestion
    was_former_foster_care: "former_foster_care_excepted",
```

jgavin-nava (approved):
LGTM otherwise

reporting-app/app/models/determination.rb
jgavin-nava: Maybe that doesnt read right was and then former am I not sure just noticed the difference between what you added

Issue #767, opened 2026-07-10
Record a mandatory exception when a former exclusion falls within the lookback


giverm:
Part of the exclusion/exception determination work under epic #732 (Data Source Hierarchy). Depends on #766 and #761. Prerequisite for #768.

## Background

With exclusion signals represented as dates (#766), a signal that is not active during the certification month but whose dates fall within the prior lookback months indicates a mandatory exception rather than an exclusion. Today `ExclusionDeterminationService` records only exclusions, so this case produces no determination. This issue records the mandatory exception for it.

Because the service now determines both exclusions and exclusion-derived exceptions, it is renamed to `ExemptionDeterminationService`, exemption being the umbrella term for exclusions and exceptions.

## User Story

As a Medicaid program processing a certification, I want a former exclusion that lapsed into the lookback recorded as a mandatory exception, so that the member is correctly excepted rather than treated as having no applicable condition.

## Acceptance Criteria

- [ ] A date-based exclusion signal that is not active during the certification month, but whose dates fall within the lookback, records a mandatory exception.
- [ ] The determination still records an exclusion when the signal is active during the certification month (unchanged from #766).
- [ ] `ExclusionDeterminationService` is renamed to `ExemptionDeterminationService`, with references and the business-process step updated accordingly.
- [ ] The appropriate event is published for the exception outcome (`DeterminedExcepted` and its not-determined counterpart), preserving existing audit-log writes.

## Out-of-scope

- Evaluating exclusion and exception conditions in a single priority-ordered pass and selecting the most durable outcome. That is the follow-on issue, #768.
- Sourcing signals from external data sources (#755, #756).

## Dependencies & Risks

- Depends on #766 (exclusion signals as dates) and #761 (return the highest-priority exclusion).
- Reuses the month-level comparison from `ExceptionDeterminationService`; the shared helper can be extracted here or in #768, coordinating with #760.

## Links & Resources

- Data Source Hierarchy tech spec, §"Exceptions from former exclusions" (linked from epic #732)
- Related: #766, #761, #757, #765 (exception pattern), #760

PR #807, opened 2026-07-23
OSCER-767: Mock Data Source

jdettmannnava:
## Ticket

Sets up #767

## Changes

- Add `Verification::Adapters::MockDrugTreatment`, a mock verification data source that derives a drug-treatment exclusion/exception outcome from the last digit of the member's VA ICN.
- Switch `VerificationDataSourcesLoader` registry validation from the Exclusion / ExternalException id registries to `Determination::REASON_CODE_MAPPING` keys (the "emit the key" convention), so a source may declare exception-fact outcomes (e.g. `:was_in_drug_treatment`) that are not exclusion ids.
- Empty `DEFAULTS` — OSCER ships no default sources; deployments register their own via the override file. Register the demo mock in `config/custom/verification_data_sources.yml`.

## Context for reviewers

- Foundation slice: nothing consults these sources yet. The mock is registered and boot-validated but inert — there is no behavior change until a later change wires `ExclusionDeterminationService` to consult data sources.
- `Verification::Adapters::VaDisabilityRating` is deliberately removed from `DEFAULTS`. It was registered but never actually consulted; dropping it keeps the follow-up service change from silently activating it. The class itself is untouched.

## Testing

- `make test args="spec/services/verification_data_sources_loader_spec.rb spec/services/verification/adapters/mock_drug_treatment_spec.rb"` → 83 examples, 0 failures.
- `make lint` clean.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/services/verification/adapters/mock_drug_treatment.rb
+++ b/reporting-app/app/services/verification/adapters/mock_drug_treatment.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+module Verification
+  module Adapters
+    # Mock data source deriving a drug-treatment outcome from the last digit of
+    # the member's va_icn:
+    #
+    #   * absent ICN                          -> :skipped
+    #   * ICN not ending in a digit           -> :success, no outcomes ("no result")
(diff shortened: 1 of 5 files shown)


jgavin-nava (approved):
LGTM

PR #808, opened 2026-07-23
OSCER-767: Consult verification data sources in exclusion determination

### LINKED ISSUE #767: Record a mandatory exception when a former exclusion falls within the lookback

jdettmannnava:
## Ticket

Resolves #767

## Changes

- `ExclusionDeterminationService` now consults the registered verification data sources after the rules engine. It starts from the rules engine's highest-priority exclusion, then walks the enabled sources by best declared priority first, calling a source only while the best exclusion it could emit could still outrank the running best (stopping early once none can). A source that emits a higher-priority exclusion wins.
- Added an exception fallback: when no exclusion applies but a source emitted an exception outcome (a non-exclusion key such as the `*_excepted` facts), the case is recorded as an exception and `DeterminedExcepted` is published.
- Added the `EXTERNAL_EXCLUSION_CHECK_STEP → DeterminedExcepted → END_STEP` transition to `CertificationBusinessProcess` so the exception fallback ends the case cleanly.

## Context for reviewers

Second and final slice of #767. The foundation slice (#807) added the mock data source, switched loader validation to emit-the-key, and left the registry inert. This slice wires the sources into the determination path; it depends on that foundation, now on `main`.

Key points in `ExclusionDeterminationService`:
- `rules_engine_best_exclusion` returns the engine's best exclusion as a `{ key:, priority: }` hash (lowest priority number wins), or `nil`.
- `consult_data_sources` sorts source candidates by `best_declared_priority` (min exclusion priority across a source's declared outcome keys; exception-only keys don't count), then gates each call with `outranks?` so a source is instantiated and called only when it could still beat the current best.
- `exclusion_priority_or_nil` is the non-raising sibling of `exclusion_priority`, used to distinguish exclusion keys from exception keys without tripping the fail-loud drift guard.

`ExclusionDeterminationService` was deliberately **not** renamed — see the exemption/exclusion/exception legal-term distinction.

Out of scope: recording *which* data source produced the winning outcome. The determination stores only the reason code, not the originating source; persisting source provenance is deferred.

## Testing

- `make test` — 2972 examples, 0 failures (line 96.68% / branch 84.15%).
- `make lint` — 553 files, no offenses.

New/expanded specs cover: rules-engine-only exclusion, a source improving on the engine's exclusion, early-stop when no source can outrank, the exception fallback path, and the new `DeterminedExcepted` transition in `certification_business_process_spec`.

Manual demo verification of the `MockDrugTreatment` source, driving the outcome via the certification's `va_icn` last digit:
- **Non-integer ICN** (no result emitted) → no exclusion or exception; case proceeds to the reporting step.
- **ICN ending in a digit divisible by 3** (emits `drug_treatment`) → case excluded.
- **ICN ending in an odd digit not divisible by 3** (emits `was_in_drug_treatment`) → case excepted (exception fallback).

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/services/exclusion_determination_service.rb
+++ b/reporting-app/app/services/exclusion_determination_service.rb
@@ -3,28 +3,108 @@
 class ExclusionDeterminationService
   include Strata::VirtualActor
   class << self
-    # Called by CertificationBusinessProcess at EXTERNAL_EXCLUSION_CHECK_STEP
-    # Service handles: evaluation, recording via model, and publishing events
-    # Business process handles: transitions and notifications
+    # Called by CertificationBusinessProcess at EXTERNAL_EXCLUSION_CHECK_STEP.
+    #
+    # Takes the highest-priority exclusion from the rules engine, then lets the
+    # verification data sources improve on it. Records an exclusion if one
+    # applies, otherwise an exception if a source emitted one, otherwise nothing.
+    #
+    # Service handles: evaluation, recording via model, and publishing events.
+    # Business process handles: transitions and notifications.
     # @param kase [CertificationCase]
     def determine(kase)
       certification = Certification.find(kase.certification_id)
-      eligibility_fact = evaluate_exclusion_eligibility(certification)
 
-      if eligibility_fact.value
-        kase.record_exclusion_determination([ highest_priority_reason_code(eligibility_fact) ], self)
-        Strata::EventManager.publish("DeterminedExcluded", { case_id: kase.id, certification_id: kase.certification_id })
+      current_best = rules_engine_best_exclusion(certification)
+      current_best, exception_keys = consult_data_sources(certification, current_best)
+
+      if current_best
+        kase.record_exclusion_determination([ reason_code(current_best[:key]) ], self)
+        publish(kase, "DeterminedExcluded")
+      elsif exception_keys.any?
+        kase.record_exception_determination([ reason_code(exception_keys.first) ], self)
+        publish(kase, "DeterminedExcepted")
       else
-        Strata::AuditLog.write!(
-          action: "case.exclusion.denied",
-          actor: self,
-          subject: certification,
-        )
-        Strata::EventManager.publish("DeterminedNotExcluded", { case_id: kase.id, certification_id: kase.certification_id })
+        Strata::AuditLog.write!(action: "case.exclusion.denied", actor: self, subject: certification)
+        publish(kase, "DeterminedNotExcluded")
       end
     end
 
     private
+
+    def publish(kase, event)
+      Strata::EventManager.publish(event, { case_id: kase.id, certification_id: kase.certification_id })
+    end
+
+    # The highest-priority exclusion (lowest priority number) the rules engine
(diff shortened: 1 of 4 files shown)


reporting-app/app/services/exclusion_determination_service.rb
```diff
@@ -3,29 +3,109 @@
...
+
+      candidates = data_sources
+        .map { |source| { source: source, priority: best_declared_priority(source) } }
+        .select { |candidate| candidate[:priority] }
+        .sort_by { |candidate| candidate[:priority] }
+
+      candidates.each do |candidate|
+        break unless outranks?(candidate, current_best)
+
+        result = candidate[:source].new.call(certification: certification)
```
jgavin-nava, line 66: Should we make sure that the results object has no errors or was skipped before continuing with the process?

jdettmannnava: Not really sure what should do on error, so I made a ticket and added a TODO in the code.

jgavin-nava: I was thinking of just logging it for now but that works for me. Hopefully have some more information or more time to think of it

jgavin-nava (approved):

PR #811, opened 2026-07-27
OSCER-809: Propagate matched data source to certification outcome

### LINKED ISSUE #809: record data source in outcome
## Background
Our outcome API needs to return the source of the data that led to a positive compliance decision. It doesn't.

## User Story
As a consumer of the output API, I want to know the source of the data that led to a positive decision.

## Acceptance Criteria
- [ ] When a data source results in a positive decision, the name of the data source is returned in the source field of the response.

## Open questions (optional)  
Should a new field be added to determination?

jdettmannnava:
## Ticket

Resolves #809

## Changes

- `ExclusionDeterminationService` names the origin of every exclusion and exception it records: the id of the verification data source that emitted the outcome, or `Determination::API_SOURCE` for one the rules engine found from API-supplied member data.
- `CertificationCase#record_exclusion_determination` takes that origin as a required argument and records it under `determination_data["data_source"]`.
- `Determination` defines `API_SOURCE` and `MEMBER_SOURCE`, and `Determination#source` resolves the reported origin — the recorded data source when present, otherwise actor provenance.
- `Api::Certifications::Outcome` reports `Determination#source` instead of branching on `automated?`.

## Context for reviewers

Follow-up to #808, which added the data source consultation but deferred recording which source produced the winning outcome.

**The origin is surfaced through the existing `outcome.source` field, not a new one.** Rules-engine and member-data outcomes keep reporting `api` / `member` exactly as before; only outcomes a verification data source produced report something new (e.g. `mock_drug_treatment`, `va_disability_rating`). One field, three possible value shapes.

**Resolution lives on `Determination#source`, not in the API projection.** `Api::Certifications::Outcome` stays a thin read of the determination. No migration — `determination_data` is jsonb.

**Worth knowing:** exclusion rows now always carry a `data_source` key, where before it was absent. Pre-existing rows don't have it and `Determination#source` falls back to actor provenance for them, so the API response is unchanged for old rows and no backfill is needed — but anything reading the raw jsonb for reporting will see two row shapes.

**Scope:** only `ExclusionDeterminationService` names its own origin so far. `ExceptionDeterminationService` and the CE compliance and exemption paths still rely on the fallback, which is why `record_exception_determination` keeps an optional `data_source:` keyword while the exclusion equivalent is required. A separate ticket moves those writers over, after which the fallback in `Determination#source` can go.

Checked for side effects: no other caller of `record_exclusion_determination`; `MemberDashboardCompliance` branches on `decision_method` so automated rows never read the jsonb Hash; no raw jsonb filters anywhere in `app/` or `lib/`; the OpenAPI spec types outcome `source` as a free-form nullable string, so no regeneration is needed; and both verification adapters set `audit_data[:source]` on every non-skipped path.

Drive-by: replaced `Determination.first` with `Determination.sole` throughout the `CertificationCase` spec, per the repo's UUID primary key guidance.

## Testing

Manually verified through the demo flow that the determination records the expected origin in each of the three recording paths:

| Outcome | Origin |
|---|---|
| `drug_treatment_excepted` | the `MockDrugTreatment` data source |
| `drug_treatment_excluded` | the `MockDrugTreatment` data source |
| `pregnancy_excluded` | the rules engine, recorded as `api` |

The last row is the behavior change called out above — a rules-engine exclusion now persists `data_source` where the key was previously absent.

```
$ make test
2984 examples, 0 failures
Line coverage: 5373 / 5557 (96.68%)
Branch coverage: 1285 / 1529 (84.04%)

$ make lint
553 files inspected, no offenses detected
```

New coverage:

- `Determination#source` — all three arms: a recorded data source, the automated fallback to `api`, and the manual fallback to `member`.
- `ExclusionDeterminationService` — records `api` for a rules-engine exclusion, the emitting source's id when a data source outranks the rules engine, and the emitting source's id for an exception; plus the real `MockDrugTreatment` adapter recording `mock_drug_treatment` for both an exclusion and an exception.
- `CertificationCase#record_exclusion_determination` — stores the given origin in `determination_data`.
- `Api::Certifications::Outcome` — reports a matched data source, and still reports `api` when none was recorded.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/api/certifications/outcome.rb
+++ b/reporting-app/app/models/api/certifications/outcome.rb
@@ -18,7 +18,7 @@ class Api::Certifications::Outcome < ValueObject
     else
       obj.status = determination.outcome
       obj.reason = determination.reasons.first
-      obj.source = determination.automated? ? "api" : "member"
+      obj.source = determination.source
     end
     obj
   end
--- a/reporting-app/app/models/certification_case.rb
+++ b/reporting-app/app/models/certification_case.rb
@@ -187,8 +187,18 @@ class CertificationCase < Strata::Case
   # Model only handles state changes - service handles selection and events
   # @param reason_codes [Array<String>] reason code(s) for the selected exclusion
   # @param actor [Strata::VirtualActor] recording the exclusion
-  def record_exclusion_determination(reason_codes, actor)
+  # @param data_source [String] origin of the exclusion — a verification data source id, or Determination::API_SOURCE
+  def record_exclusion_determination(reason_codes, actor, data_source)
     certification = Certification.find(certification_id)
+
+    # determination_data is a jsonb column holding a structured Hash (OSCER convention; see the
+    # Determination docs). For automated exclusions the type is carried by the reason codes, which
+    # the dashboard reads via decision_method branching, so we record those codes here. It must stay
+    # a Hash: writing a JSON String (e.g. reasons.to_json) double-encodes into the jsonb column and
+    # reads back as a String, which 500'd the member dashboard. See
+    # https://github.com/navapbc/oscer/issues/680. Api::Certifications::Outcome surfaces
+    # "data_source" as the outcome source.
+    determination_data = { "exclusion_reasons" => reason_codes, "data_source" => data_source }
 
     transaction do
       self.exemption_request_approval_status = "approved"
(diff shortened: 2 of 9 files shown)


giverm (approved):
Looks good!

Issue #805, opened 2026-07-22
Orchestrate non-exclusion verification data sources in priority order


giverm:
## Background

Build the orchestrator the Data Source registry (#756) was designed for but deferred. Per the Data
Source Hierarchy spec (#732, "Add non-Exclusions to the Data Source Registry"): a new `system_process`
after `external_community_engagement_check` that calls ordered non-exclusion sources, stops on the
first positive result, and takes over the negative-determination + `report_activities` handoff from
the community-engagement check.

No real non-exclusion source exists yet, so this is orchestrator-only, proven against a mock source
behind a feature flag (off in production, so behavior is unchanged).

## Acceptance Criteria

- [ ] New `system_process` after `external_community_engagement_check`; iterates enabled sources with
      an `order` (ascending), stops on first `:success`. `order: nil` sources are skipped.
- [ ] Records the producing source id for the API `outcome`; `:skipped`/`:error` do not stop the pass.
- [ ] Owns the negative-determination + `report_activities` trigger when no source succeeds; the
      community-engagement check routes not-met members into the orchestrator.
- [ ] Mock `Verification::DataSource` registered only when `FEATURE_MOCK_DATA_SOURCES` is on; add that
      flag to `Features::FEATURE_FLAGS` (default false).
- [ ] Automated tests via mock adapters: ordering, stop-on-success, skip/error, empty-set, provenance.

## Out-of-scope

- Exclusion ranking (`Exclusion.priority_order`).
- Migrating existing exception/CE determinations into registered adapters.
- Hybrid (exclusion + non-exclusion) sources.

## Links

- Epic #732; registry #755/#756; related #767/#768

giverm:
Completed and merged. The orchestrator is load-bearing on `main` as of 2026-07-31.

## Delivered by

| PR | Commit | What it added |
|----|--------|---------------|
| #812 | `cfbd8fdb` | `Verification::DataSourceOrchestrator` (stop-on-first by `order`), `OrchestrationResult`, `MockEmergencyCounty` |
| #823 | `e58b0399` | Outcome categorization and boot-time validation of the registry |
| #828 | `2eb455d9` | Extracted the shared in-hand community-engagement assessment |
| #829 | `b016b061` | Registered `MockCommunityEngagement` |
| #822 | `13cac2d7` | The wiring: `verification_data_source_check` consults sources before concluding non-compliance |

Full suite on merged `main`: 3163 examples, 0 failures, 96.76% line / 84.22% branch.

[...]

PR #812, opened 2026-07-27
Orchestrate non-exclusion verification data sources in priority order

giverm:
## Ticket

Part of #805. The business-process wiring that consumes the orchestrator is a separate follow-up PR.

## Changes

- Add `Verification::DataSourceOrchestrator`: evaluates the order-bearing (non-exclusion) verification data sources for a certification in ascending `order`, stopping at the first result that is `success?` and carries at least one outcome.
- Add `Verification::OrchestrationResult` value object: carries `satisfied`, the winning `source_id` and its full `DataSourceResult`, plus `attempted` (every source called, with its result) so skipped or errored sources are never silently dropped.
- Add `Verification::Adapters::MockEmergencyCounty`: a demo non-exclusion source emitting `resides_in_declared_emergency_county`, derived from the last digit of the member's `va_icn`.
- Register `mock_emergency_county` in `config/custom/verification_data_sources.yml` with `order: 10`.

## Context for reviewers

The verification Data Source registry (#756) stores each source's call `order` and states that consumers sort by `order` themselves, but nothing consumed it. This adds that consumer.

Nothing calls the orchestrator yet, so this makes no behavior change on its own; the service that consumes it is a separate follow-up PR. This follows the same pattern as #807, which registered its mock before #808 wired it into determination. Registering `mock_emergency_county` as `enabled: true` is safe even though nothing consumes the orchestrator: it declares only a non-exclusion outcome, so `ExclusionDeterminationService` (which selects sources by exclusion priority) skips it, and only the orchestrator picks up order-bearing sources.

**Note on scope:** #805's acceptance criteria mention a `FEATURE_MOCK_DATA_SOURCES` flag. That AC predates #807, which established registering mocks directly in the deployment-owned `config/custom/verification_data_sources.yml` (gated by `enabled:`) with no flag. This PR follows that convention, so no flag is added; the ticket AC is left for follow-up.

## Testing

- `make test` (full suite): 3031 examples, 0 failures (96.71% line / 84.12% branch coverage).
- `make lint`: no offenses.
- Specs cover the orchestrator's ordering, stop-on-first-positive, skip of `nil`-order and disabled sources, empty-set handling, and fail-loud propagation of unexpected errors (unit, stubbed registry), plus a real-registry integration pass against the booted config; the mock adapter's every branch; and loader registration and boot-validation of the new source.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/verification/orchestration_result.rb
+++ b/reporting-app/app/models/verification/orchestration_result.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Verification
+  # Result of one orchestration pass over the ordered non-exclusion data sources
+  # (see {Verification::DataSourceOrchestrator}).
+  #
+  # It carries enough for a caller to both make a determination and audit the
+  # pass: whether a source produced a positive result (+satisfied+), the winning
+  # +source_id+ and its full {DataSourceResult} (which includes +outcomes+ and
+  # redacted +audit_data+), and +attempted+ — the ordered list of every source
+  # called, each as +{ source_id:, result: }+ — so skipped/errored sources are
+  # never silently dropped.
+  #
+  # Value object in the style of {DataSourceResult}.
+  class OrchestrationResult < ValueObject
+    attribute :satisfied, default: false
+    attribute :source_id
+    attribute :result
+    attribute :attempted, default: -> { [] }
+
+    def satisfied?
+      satisfied
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


jdettmannnava (approved):
It feels a little overengineered to me, but that will likely change when I see the next PR.

Also, it would be nice to have a mock that emitted a successful community engagement response (e.g. hours met), as it calls a different certification case endpoint.

Issue #810, opened 2026-07-24
Handle Errors on Data Source Failure


jdettmannnava:
## Background
When a data source fails, we simply roll on to the next data source. Unsure if this should change.

## User Story

As a developer, I want a clear rule on what to do when a data source fails so that exclusion checks run well.

## Acceptance Criteria

- [ ] On data source error in ExclusionDeterminationService, the service TBD

## Open questions (optional)  
What the data source should do.

Issue #755, opened 2026-07-07
Refactor VA disability integration into the verification adapter contract


jgavin-nava:
**Ticket B of 3** in the generic verification data-source adapter slice. Part of epic #732 (Data Source Hierarchy). Depends on A (contract + result object, #754).

## Background

`VeteranDisabilityService` is the first external verification data source and will become the reference implementation of the new `Verification::DataSource` contract (ticket A). Today it exposes a domain-specific method returning raw VA JSON or `nil`, and the 100%-rating interpretation lives in `Rules::ExclusionRuleset#is_veteran_with_disability`. This ticket makes the VA path conform to the contract: uniform entry point, flat outcome symbols, redacted audit data, explicit error/skip handling.

The HTTP transport layer (`VeteranAffairsAdapter < DataIntegration::BaseAdapter`) stays as-is — this ticket only touches the domain/service layer that wraps it.

## User Story

As an OSCER engineer, I want the VA disability integration to implement the verification adapter contract so that it returns outcome symbols + audit data like every other data source, owns its ICN precondition, and normalizes transport failures instead of returning `nil`.

## Acceptance Criteria

- [ ] A VA data source (`VeteranDisabilityService` conformed, or a new `Verification::Adapters::VaDisabilityRating`) implements `#call(certification:)` and returns a `Verification::DataSourceResult`.
- [ ] Emits `:is_veteran_with_disability` when the VA combined disability rating == 100; returns `:success` with empty `outcomes` otherwise.
- [ ] Owns its precondition: when `certification.member_data&.va_icn` is blank, returns `status: :skipped` (no HTTP call). Removes the need for the caller to null-check ICN first.
- [ ] Transport / token failures (`VeteranAffairsAdapter::ApiError`, `VaTokenManager::TokenError`) are caught and returned as `status: :error` — never raised, never `nil`.
- [ ] `audit_data` contains a redacted summary (e.g. combined rating %, request id if available) — not the full VA payload.
- [ ] The 100%-threshold interpretation moves into the adapter; `Rules::ExclusionRuleset#is_veteran_with_disability` JSON parsing is left untouched for now (its removal is downstream/out of scope) — call out the temporary duplication explicitly in the PR.
- [ ] Existing VA specs are ported to result-object assertions: success+match, success+no-match, missing ICN → skipped, `ApiError`/`TokenError` → error (with `audit_data` present).
- [ ] Includes the shared contract examples from ticket A.

## Out-of-scope

[...]

Issue #732, opened 2026-07-01
Epic: Data Source Hierarchy


jdettmannnava:
We need to be able to configure the order and conditions under which external data sources are called.

[Product Brief](https://docs.google.com/document/d/1cBRWifOIFgVJJpx6mVEBnEa9a5sczwopqWjp7NAfxtQ/edit?usp=sharing)

PR #822, opened 2026-07-29
Consult verification data sources before concluding non-compliance

giverm:
## Ticket

Part of #805

Last of four in a stack: #823, #828, #829, this. Review in that order; each merge retargets this one
automatically.

The orchestrator engine, result object and first mock shipped in #812. This is the wiring that makes
them load-bearing.

## Changes

Flow:

- Add `verification_data_source_check` as a trailing determination step, running
  `Verification::DataSourceOrchestrator` over the registry's order-bearing sources
  (`DataSourceCheckService`).
- Re-point the community-engagement check's not-met branch into that step via a new
  `DeterminedCommunityEngagementNotMet` event, which has no `NotificationsEventListener` subscription.
- Move the negative determination and the `Insufficient` / `ActionRequired` split out of
  `CommunityEngagementCheckService` and into the new step.

Determinations:

- Add `CertificationCase#record_data_source_ce_determination` and
  `Determination::CALCULATION_TYPE_DATA_SOURCE_CE` for source-attested community engagement, rather
  than reusing `record_external_ce_combined_assessment`, whose value object serializes an in-hand hours
  and income assessment a source does not report.
- Dispatch on #823's outcome categories to pick the determination an emitted outcome calls for, raising
  `UnsupportedOutcomeError` on an uncategorized one.
- Record **one** reason code for an exception and **all** of them for community engagement. Each follows
  its own determination shape's precedent: `ExclusionDeterminationService` and
  `ExceptionDeterminationService` both take the first match, while
  `external_ce_combined_reason_codes` records both codes when both tracks pass. Whichever key the
  source emitted first wins.

Specs:

- Add `certification_business_process_reachability_spec`, which drives the real cascade through the real
  registry with nothing in the determination path stubbed and asserts over the live registry that every
  enabled order-bearing source is creditable, so a future source that starves an existing one fails
  there.
- Move `latest_determination_for` and the cascade bootstrap stub into `spec/support/`.

## Context for reviewers

**Sequencing.** In-hand checks run first (exclusion, exception, community engagement, all against data
already in OSCER), then the outbound pass, then exactly one determination once all the evidence is in.
That is why the negative moved here: recording non-compliance at the community-engagement check and then
letting a source except the member would leave a superseded `not_compliant` row and a false
`case.activity_report.denied` audit line. Plugging in a real source should be registration only, giving
it an `order` and declaring its outcomes.

**Where it sits.** #732 configures the order and conditions under which external sources are called:
contract #754, VA adapter #755, registry #756, exclusion-path consultation #808, API provenance #811,
ordering engine #812. Nothing consumed #812 until now.

**Open seams.** Proven against mocks only, since no real non-exclusion source exists, which is why #805
is scoped orchestrator-only. Cross-source aggregation stays deferred until a concrete source defines it.
No `ExternalException.enabled?` gate on the exception path: a deployment owns both its sources and its
`external_exceptions.yml`, so one that disables an exception would not ship a source attesting it. Note
that `mock_emergency_county` does emit an optional exception today, so the OSCER-shipped mocks are the
one way to construct that mismatch by config alone.

**Error handling still needs a ticket**, narrower than it looked. #810's acceptance criterion is still
`TBD` in the issue, but the design question was answered in `#tss-oscer` on 2026-07-27: continue through
the remaining sources marking the failed one as errored, and distinguish "checked everything, found no
compliance" from "checked everything but one source errored, so this is not comprehensive". Today an
errored source yields a member-facing `not_compliant` indistinguishable from a clean negative, evidenced
only by `log_failed_attempts`. Unreachable while every registered source is an in-memory mock that
cannot raise. Two gaps remain unticketed and share one root cause, that this trailing step has no error
boundary: an undeclared error propagates into Strata's `rescue Exception` and strands the case, and an
order-bearing hybrid source stalls the same silent way.

## Testing

- Suite green: 3163 examples, 0 failures. 96.76% line, 84.22% branch.
- RuboCop clean across 577 files.
- Ran in development against the shipped config, driving certifications through the real cascade from
  creation and inspecting the resulting case step and determination row. All five documented paths
  behaved as the registry describes, each writing exactly one determination:

  | `va_icn` last char | email has `ce-met` | Result |
  |---|---|---|
  | 0, 3, 6, 9 | any | `excluded` (mock_drug_treatment) |
  | 1, 5, 7 | any | `excepted` (mock_drug_treatment) |
  | 2, 4, 8 | any | `excepted` (mock_emergency_county) |
  | not a digit, or absent | no | `not_compliant`, routed to `report_activities` |
  | not a digit, or absent | yes | `compliant` (mock_community_engagement) |

- That walkthrough is what surfaced an unreachable community-engagement branch; the matrix was
  re-verified after the fix.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/certification_case.rb
+++ b/reporting-app/app/models/certification_case.rb
@@ -282,10 +282,9 @@ class CertificationCase < Strata::Case
     )
   end
 
-  # Called by DataSourceCheckService when an outbound source attests community engagement is met.
-  # NOT +#record_external_ce_combined_assessment+: that shape records the *in-hand* assessment, and a
-  # source attests without reporting hours, so +hours_ok: true+ there would claim in-hand support the
-  # data does not give. Mirrors +#record_exception_determination+ instead, with a flat payload.
+  # Called by DataSourceCheckService when a data source attests community engagement is met. A source
+  # returns a verdict, not hours or income, so this cannot reuse
+  # +#record_external_ce_combined_assessment+, whose payload states figures we would have to invent.
   def record_data_source_ce_determination(reason_codes, actor, data_source:)
     # jsonb column; must stay a Hash, not a JSON String (see record_exclusion_determination).
     determination_data = {
--- a/reporting-app/app/models/certification_case.rb
+++ b/reporting-app/app/models/certification_case.rb
@@ -282,6 +282,27 @@ class CertificationCase < Strata::Case
     )
   end
 
+  # Called by DataSourceCheckService when an outbound source attests community engagement is met.
+  # NOT +#record_external_ce_combined_assessment+: that shape records the *in-hand* assessment, and a
+  # source attests without reporting hours, so +hours_ok: true+ there would claim in-hand support the
+  # data does not give. Mirrors +#record_exception_determination+ instead, with a flat payload.
+  def record_data_source_ce_determination(reason_codes, actor, data_source:)
+    # jsonb column; must stay a Hash, not a JSON String (see record_exclusion_determination).
+    determination_data = {
+      "calculation_type" => Determination::CALCULATION_TYPE_DATA_SOURCE_CE,
+      "data_source" => data_source
+    }
+
+    # Reused for its close_on_compliant default, so a source-attested compliant member leaves the
+    # open-case queues exactly as an in-hand compliant member does.
+    record_automated_ce_compliance(
+      :compliant,
+      determination_data,
+      reasons: reason_codes,
+      actor: actor
+    )
+  end
+
   # External CE check: one automated determination with both tracks in +determination_data+.
   # Member is compliant if either +hours_ok+ or +income_ok+; not compliant only when both are false.
   # Events/notifications are published by CommunityEngagementCheckService (via Strata).
--- a/reporting-app/app/services/data_source_check_service.rb
+++ b/reporting-app/app/services/data_source_check_service.rb
@@ -1,0 +1,123 @@
+# frozen_string_literal: true
+
+# Called by CertificationBusinessProcess at VERIFICATION_DATA_SOURCE_CHECK_STEP (OSCER-805).
+# Consults the ordered non-exclusion data sources via {Verification::DataSourceOrchestrator} and
+# records the winning outcome, or concludes non-compliance and hands the member to
+# report_activities. This is where OSCER calls OUT; the preceding steps assess data already in hand.
+#
+# The negative is recorded HERE, not at the community-engagement check: a member a source goes on to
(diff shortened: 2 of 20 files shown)


reporting-app/app/services/data_source_check_service.rb
```diff
@@ -0,0 +1,123 @@
...
+    end
+
+    def record_source_attestation(kase, orchestration)
+      outcomes = orchestration.result.outcomes
+      validate_outcomes!(outcomes, orchestration.source_id)
+      data_source = orchestration.source_id.to_s
+
+      exception_keys = outcomes & Determination::EXCEPTION_OUTCOME_KEYS
+
+      # KNOWN GAP: no ExternalException.enabled? gate, so a source can attest an optional exception a
```
jdettmannnava: This comment confuses me.

reporting-app/app/services/data_source_check_service.rb
```diff
@@ -0,0 +1,123 @@
...
+      # deployment disabled, which the in-hand check honors (exception_determination_service.rb:151).
+      # Unreachable while external_exceptions.yml carries no overrides. Do NOT gate it here: the
+      # orchestrator already stopped at the first source to return anything, so dropping a disabled
+      # outcome would shadow a valid one from a higher-order source. Belongs in
+      # DataSourceOrchestrator#positive?, with an explicit outcome-key to ExternalException-id map
+      # (medical_travel vs traveling_for_medical_care).
+      #
+      # An exception outranks a CE attestation from the same source, mirroring the flow's own order.
+      if exception_keys.any?
+        kase.record_exception_determination(reason_codes(exception_keys), self, data_source: data_source)
```
jdettmannnava: Although this is an array, we are only ever returning one reason. See ExclusionDeterminationService

reporting-app/app/models/certification_case.rb
```diff
@@ -282,6 +282,27 @@ def record_income_compliance(outcome, income_data, close_on_compliant: true)
     )
   end
 
+  # Called by DataSourceCheckService when an outbound source attests community engagement is met.
+  # NOT +#record_external_ce_combined_assessment+: that shape records the *in-hand* assessment, and a
```
jdettmannnava: I don't really understand this comment

reporting-app/app/services/data_source_check_service.rb
jdettmannnava: Also, at this point we shouldn't have to worry about data sources emitting disabled exceptions, as all data sources are custom, and if the client does not want an exception, the data source should not emit it.

jdettmannnava (approved):
Some small recommendations

PR #823, opened 2026-07-29
Enforce the verification data source outcome contract

giverm:
## Ticket

Part of #805

First of four in a stack: this, #828, #829, #822. Lands inert; #822 is the consumer.

## Changes

- Split `Determination::REASON_CODE_MAPPING` into six groups named for the outcome that records them
  (exclusion, exception, CE met, CE insufficient, exemption request, denial response), and build the
  flat lookup by merging them, so every existing caller is untouched.
- Derive `EXCEPTION_OUTCOME_KEYS`, `CE_OUTCOME_KEYS`, `NON_EXCLUSION_OUTCOME_KEYS` and
  `DENIAL_RESPONSE_REASONS` from their groups rather than restating their keys.
- Boot-validate in `VerificationDataSourcesLoader` that an order-bearing source declares only
  categorized outcomes. Exclusion keys stay permitted; `order: nil` sources are skipped.
- Add four `determination_spec` guards: the merged size equals the sum of the groups, CE met and
  insufficient codes stay distinguishable, no exclusion key reaches the non-exclusion partition, and
  the exception group's size matches `ExceptionDeterminationService::EXCEPTION_CHECKS`.

## Context for reviewers

`REASON_CODE_MAPPING` spans exclusion, exception, community-engagement, exemption and denial-response
codes in one namespace, and nothing said which category an emitted outcome falls in. Grouping it by
recording outcome makes a code's category follow from where it is defined, so a consumer can pick a
determination shape without a separate list of keys to keep in step. `EXEMPTION_REASONS` stays a
`values_at` view because it deliberately spans groups: three exclusion conditions also stand as
exemption reasons.

Boot validation rather than a spec, because an uncategorized outcome is a real `REASON_CODE_MAPPING`
key: it clears the existing `validate_outcome_ids!` check and fails only at determination time, inside
Strata's `rescue Exception` plus a log, stranding the case with no determination, notification or staff
task. `config/custom/verification_data_sources.yml` is also deployment-owned, so a registration our CI
never sees can introduce one.

The four guards cover what grouping makes silent. `reduce(:merge)` keeps the last value for a key
defined twice, so the size check catches a duplicate the merge would swallow. A code filed under the
wrong group preserves the total, so the suffix and exclusion-key guards cover mis-categorization. The
consequential cases are a CE insufficient code recorded as `:compliant`, and an exclusion key reaching
a pass with no determination shape for one. The exception group cannot be derived from
`EXCEPTION_CHECKS`, which holds check method names (`pregnancy`, `medical_travel`) rather than outcome
keys (`was_pregnant`, `traveling_for_medical_care`), so a size assertion is the available link.

This does not change the exclusion path: the boot check returns early for `order: nil` sources, which
is how exclusion-only sources are registered.

## Testing

- Suite green on this branch alone: 3052 examples, 0 failures. 96.72% line, 84.05% branch.
- RuboCop clean across 569 files.
- Boot validation exercised against the real registry (the loader spec loads the shipped config and
  asserts it passes), plus a stubbed source declaring `:exemption_request_compliant`, a real key in no
  category, rejected at boot.
- Each new guard confirmed to fail on the mis-categorization it describes, not just to pass today.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/determination.rb
+++ b/reporting-app/app/models/determination.rb
@@ -66,7 +66,11 @@ class Determination < Strata::Determination
   API_SOURCE = "api"
   MEMBER_SOURCE = "member"
 
-  REASON_CODE_MAPPING = {
+  # REASON_CODE_MAPPING below merges these groups into the flat lookup every caller uses. A new code
+  # goes in one group and nowhere else; its category follows from where it is defined.
+
+  # Recorded as +:excluded+.
+  EXCLUSION_REASON_CODES = {
     age_under_19: "age_under_19_excluded",
     age_over_65: "age_over_65_excluded",
     is_pregnant: "pregnancy_excluded",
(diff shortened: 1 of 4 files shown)


reporting-app/app/models/determination.rb
```diff
@@ -103,6 +103,33 @@ class Determination < Strata::Determination
     participating_in_other_program: "other_program_excepted"
   }.freeze
 
+  # REASON_CODE_MAPPING is one flat namespace; these lists partition its non-exclusion keys so a
+  # consumer can tell which determination an outcome calls for. VerificationDataSourcesLoader
+  # boot-validates the partition against every order-bearing source's declared outcomes.
+
+  # Attests a non-exclusion exception applies. Recorded as +:excepted+.
+  EXCEPTION_OUTCOME_KEYS = %i[
```
jdettmannnava: I'm not a big fan of the duplication. Would it be better (and I could easily be convinced otherwise) of having a reason_code mapping for each outcome type, then merge them for lookup?

jdettmannnava (approved):

PR #828, opened 2026-07-30
Extract the shared in-hand community-engagement assessment

giverm:
## Ticket

Part of #805

Second of four in a stack: #823, this, #829, #822. A pure refactor, split out so it can be verified as
a no-op on its own.

## Changes

- Extract the in-hand community-engagement derivation from `CommunityEngagementCheckService#determine`
  into a public `.assess`, returning an `Assessment` struct: both aggregates, both per-track verdicts,
  and `#met?`.
- Fold the private `hours_compliant?` wrapper into `.assess`.
- Add a direct `.assess` spec covering hours-only, income-only and neither, so `met?` is exercised
  through both sides of its disjunction rather than only the hours track.

## Context for reviewers

The trailing data-source step in #822 has to recompute this assessment when no source produced an
outcome, because Strata hands a `system_process` only the case and never the triggering event.
Deriving the same four values in both steps would let them disagree about the same certification, so
the derivation lives in one place.

Split from #822 because the extraction and the behavioural change that consumes it are hard to review
together: the diff shows a rewritten method body with no easy way to tell which lines moved and which
changed behaviour. Reviewed alone the evidence is simple, since every pre-existing example passes with
the spec file unmodified. `.determine` still records the combined determination and still publishes all
three community-engagement events; #822 changes that.

## Testing

- Suite green on this branch alone: 3054 examples, 0 failures. 96.72% line, 84.05% branch. Branch
  coverage is unchanged from #823, as expected for a refactor over already-covered paths.
- RuboCop clean across 569 files.
- The 16 pre-existing `community_engagement_check_service_spec` examples pass with no edits to the spec
  file, which is the substance of the no-op claim. Three new examples cover `.assess` and `#met?`.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/spec/services/community_engagement_check_service_spec.rb
+++ b/reporting-app/spec/services/community_engagement_check_service_spec.rb
@@ -38,27 +38,46 @@ RSpec.describe CommunityEngagementCheckService do
   # .assess is the derivation .determine used to inline. It is public so a second step can reuse it
   # rather than re-deriving the same four values and risking a different verdict.
   describe ".assess" do
-    it "returns both aggregates and their per-track verdicts" do
-      create_external_hourly_activity_for(certification, hours: 85)
+    let(:over_hours) { HoursComplianceDeterminationService::TARGET_HOURS + 5 }
+    let(:under_hours) { HoursComplianceDeterminationService::TARGET_HOURS / 2 }
+    let(:over_income) { IncomeComplianceDeterminationService::TARGET_INCOME_MONTHLY + 20 }
+    let(:under_income) { IncomeComplianceDeterminationService::TARGET_INCOME_MONTHLY / 2 }
+
+    it "returns both aggregates and their per-track verdicts when only hours pass" do
+      create_external_hourly_activity_for(certification, hours: over_hours)
+
+      assessment = described_class.assess(certification)
+
+      expect(assessment.hours_data[:total_hours]).to eq(over_hours)
+      expect(assessment.income_data[:total_income]).to be_zero
+      expect(assessment.hours_ok).to be(true)
+      expect(assessment.income_ok).to be(false)
+      expect(assessment.met?).to be(true)
+    end
+
+    # met? is hours_ok || income_ok; without this the income side is never exercised.
+    it "reports met? when only the income track passes" do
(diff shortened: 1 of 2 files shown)


reporting-app/spec/services/community_engagement_check_service_spec.rb
```diff
@@ -35,6 +35,36 @@ def latest_determination_for(certification_id)
...
+
+      assessment = described_class.assess(certification)
+
+      expect(assessment.hours_data[:total_hours]).to be_positive
+      expect(assessment.income_data).to have_key(:total_income)
+      expect(assessment.hours_ok).to be(true)
+      expect(assessment.income_ok).to be(false)
+    end
+
+    it "reports met? when either track passes" do
```
jdettmannnava: This only tests hours?

reporting-app/spec/services/community_engagement_check_service_spec.rb
```diff
@@ -35,6 +35,36 @@ def latest_determination_for(certification_id)
...
+    it "returns both aggregates and their per-track verdicts" do
+      create_external_hourly_activity_for(certification, hours: 85)
+
+      assessment = described_class.assess(certification)
+
+      expect(assessment.hours_data[:total_hours]).to be_positive
+      expect(assessment.income_data).to have_key(:total_income)
+      expect(assessment.hours_ok).to be(true)
+      expect(assessment.income_ok).to be(false)
+    end
```
jdettmannnava, line 56: Should this also test met?

reporting-app/spec/services/community_engagement_check_service_spec.rb
```diff
@@ -35,6 +35,36 @@ def latest_determination_for(certification_id)
...
+
+    it "reports met? when either track passes" do
+      create_external_hourly_activity_for(certification, hours: 85)
+
+      expect(described_class.assess(certification).met?).to be(true)
+    end
+
+    it "reports met? false when neither track passes" do
+      create_external_hourly_activity_for(certification, hours: 1)
```
jdettmannnava: This only tests hours?

jdettmannnava (approved):
I thought a couple additional tests might be good

PR #829, opened 2026-07-30
Register the community-engagement mock data source

giverm:
## Ticket

Part of #805

Third of four in a stack: #823, #828, this, #822. Lands inert; #822 is the consumer.

## Changes

- Add `Verification::Adapters::MockCommunityEngagement`, which attests community engagement when the
  member's email contains `ce-met`.
- Register it in the deployment-owned `config/custom/verification_data_sources.yml` at `order: 20`,
  behind `mock_emergency_county` at `order: 10`.
- Extract `stub_registry` / `registry_entry` into `VerificationRegistryHelpers`, now that a second spec
  stubs the booted registry.
- Update `data_source_orchestrator_spec`'s real-registry cases for a second order-bearing source.

## Context for reviewers

Inert, the same staging #812 used for `MockEmergencyCounty`: `DataSourceOrchestrator` still has no
consumers. The adapter declares no exclusion outcome, so `ExclusionDeterminationService` drops it
(`best_declared_priority` is nil) and only the orchestrator's pass will ever consult it.

The trigger is email rather than `va_icn` because `MockDrugTreatment` runs at the earlier exclusion step
and terminates the case for seven of ten `va_icn` last digits, leaving only even ones, which are
`MockEmergencyCounty`'s. A `va_icn`-keyed trigger here would be starved before it ran. Email is safe
because no determination path reads it, and the substring idiom follows `Auth::MockAdapter`'s
"unconfirmed"/"mfa". A real source must not copy it, and the adapter says so: email both describes the
member and joins them to their certification (`Certification.find_by_member_email`), so a single
member-influenceable attribute would select the record and drive a favourable outcome.

The orchestrator spec churns because a third order-bearing source changes the `attempted` list its
real-registry cases assert. The two mocks read different member fields, so an even-digit `va_icn` must
stop at `order: 10` without consulting the CE source; that precedence is now pinned by a spec rather
than left implicit in registry order.

## Testing

- Suite green on this branch alone: 3107 examples, 0 failures. 96.73% line, 84.14% branch.
- RuboCop clean across 572 files.
- Registration proven against the real booted registry rather than a stub: the loader spec asserts both
  mocks are enabled, order-bearing, and ordered emergency-county first.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/spec/services/verification/adapters/mock_community_engagement_spec.rb
+++ b/reporting-app/spec/services/verification/adapters/mock_community_engagement_spec.rb
@@ -1,0 +1,112 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Verification::Adapters::MockCommunityEngagement do
+  subject(:result) { described_class.new.call(certification: certification) }
+
+  let(:certification) do
+    build(:certification, member_data: build(:certification_member_data, **member_data_attrs))
+  end
+  let(:member_data_attrs) { { account_email: account_email } }
+  let(:trigger) { described_class::TRIGGER_EMAIL_SUBSTRING }
+
+  describe ".declared_outcomes" do
+    let(:account_email) { nil }
+
+    it "declares the hours-met community-engagement outcome key" do
+      expect(described_class.declared_outcomes).to contain_exactly(:hours_reported_compliant)
+    end
+
+    it "declares only Determination reason-code mapping keys" do
+      expect(described_class.declared_outcomes).to all(be_in(Determination::REASON_CODE_MAPPING.keys))
+    end
+
+    it "declares only non-exclusion (order-bearing) outcome keys" do
+      expect(described_class.declared_outcomes).to all(satisfy { |key| Exclusion.find(key).nil? })
+    end
+  end
+
+  describe "#call" do
+    context "when the member has no email" do
+      let(:account_email) { nil }
+
+      it_behaves_like "a skipped verification result"
+    end
+
+    context "when the email is a blank string" do
+      let(:account_email) { "" }
+
+      it_behaves_like "a skipped verification result"
+    end
+
+    context "when the email does not contain the trigger" do
+      let(:account_email) { "member@example.com" }
+
+      it_behaves_like "a successful verification result"
+
+      it "returns no result (empty outcomes)" do
+        expect(result.outcomes).to eq([])
+      end
+
+      it "tags the audit data with the source" do
+        expect(result.audit_data[:source]).to eq("mock_community_engagement")
+      end
+    end
+
+    context "when the email contains the trigger" do
+      let(:account_email) { "member+#{described_class::TRIGGER_EMAIL_SUBSTRING}@example.com" }
+
+      it_behaves_like "a successful verification result"
+
+      it "attests the community-engagement requirement is met" do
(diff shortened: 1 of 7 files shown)


reporting-app/spec/services/verification/adapters/mock_community_engagement_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+require "rails_helper"
+
+RSpec.describe Verification::Adapters::MockCommunityEngagement do
+  subject(:result) { described_class.new.call(certification: certification) }
+
+  let(:certification) do
+    build(:certification, member_data: build(:certification_member_data, **member_data_attrs))
+  end
+  let(:member_data_attrs) { { account_email: account_email } }
+  let(:trigger) { described_class::TRIGGER_EMAIL_SUBSTRING }
```
jdettmannnava, line 12: Shouldn't you be using this?

reporting-app/spec/services/verification/adapters/mock_community_engagement_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+        expect(result.outcomes).to eq([])
+      end
+
+      it "tags the audit data with the source" do
+        expect(result.audit_data[:source]).to eq("mock_community_engagement")
+      end
+    end
+
+    context "when the email contains the trigger" do
+      let(:account_email) { "member+#{described_class::TRIGGER_EMAIL_SUBSTRING}@example.com" }
```
jdettmannnava:
```suggestion
      let(:account_email) { "member+#{trigger}@example.com" }
```

reporting-app/spec/services/verification/adapters/mock_community_engagement_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    context "when the trigger appears in the local part in mixed case" do
+      let(:account_email) { "Member.CE-Met@Example.com" }
+
+      it "matches case-insensitively so a capitalized demo email still fires" do
+        expect(result.outcomes).to eq([ :hours_reported_compliant ])
+      end
+    end
+
+    context "when the trigger is present only on the contact email" do
+      let(:member_data_attrs) { { contact: { email: "contact+#{described_class::TRIGGER_EMAIL_SUBSTRING}@example.com" } } }
```
jdettmannnava:
```suggestion
      let(:member_data_attrs) { { contact: { email: "contact+#{trigger}@example.com" } } }
```

reporting-app/spec/services/verification/adapters/mock_community_engagement_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+        expect(result.outcomes).to eq([ :hours_reported_compliant ])
+      end
+
+      it "tags the audit data with the source" do
+        expect(result.audit_data[:source]).to eq("mock_community_engagement")
+      end
+    end
+
+    context "when the trigger appears in the local part in mixed case" do
+      let(:account_email) { "Member.CE-Met@Example.com" }
```
jdettmannnava:
```suggestion
      let(:account_email) { "Member.#{trigger.camelize}@Example.com" }
```

jdettmannnava (approved):
Some test cleanup recommended

PR #831, opened 2026-07-31
OSCER-813: Use new API member data to evaluate exclusions

### LINKED ISSUE #813: Update API
## Background
The API as it stands needs revision to accommodate more robust checking practices. We have a [tech spec](https://docs.google.com/document/d/1cBEzUPhfUSJxL6NSnCTFa5-kj-iwStSLoUDGJLLzPEY/edit?usp=sharing).

## User Story
As an API consumer, I want an API that is more robust so that I can get as many external checks done as possible.

## Acceptance Criteria

- [ ] API accepts the inputs outlined in the tech spec
- [ ] Member data updated correctly trigger exemptions, exceptions, and community-service complicance
[Tech spec](https://docs.google.com/document/d/1cBEzUPhfUSJxL6NSnCTFa5-kj-iwStSLoUDGJLLzPEY/edit?usp=sharing)

jdettmannnava:
## Ticket

Resolves #813

## Changes

- Demo CreateForm adds API-style exemptions to member data
- ExclusionRuleset uses MemberData::Exemption objects to determine exclusions
- ExclusionDeterminationService extracts exemptions from member data to set ruleset facts
- member data exemption factory added

## Context for reviewers

The new API will be passing in different data, and exclusions need to be determined based on this data rather than the old API.
The demo form had to be updated to add the exemptions correctly for processing.

## Testing

Automated testing.

♻️ Environment destroyed ♻️


reporting-app/app/forms/demo/certifications/create_form.rb
```diff
@@ -114,6 +129,19 @@ def apply_external_exception(member_data, months)
           member_data.dates_participating_in_other_program = months
         end
       end
+      def end_exemption(exemption_type, cert_date)
+        {
+          type: exemption_type,
+          valye: true,
```
jgavin-nava:
```suggestion
          value: true,
```

reporting-app/app/models/rules/exclusion_ruleset.rb
```diff
@@ -17,75 +17,67 @@ class ExclusionRuleset < Strata::Rules::MedicaidRuleset
...
       return if certification_date.nil?
+      return if incarceration.nil?
 
       cert_month = certification_date.beginning_of_month
-      Array(dates_incarcerated).any? do |date|
-        window_start = date.beginning_of_month
-        window_start <= cert_month && cert_month <= window_start + INMATE_BUFFER_MONTHS.months
+      Array(incarceration.periods).any? do |period|
+        next unless period.period_end
+        cert_month <= period.period_end.end_of_month + INMATE_BUFFER_MONTHS.months
```
jgavin-nava: Do we need to check if the cert_month is before the period.period_start as well?

reporting-app/app/models/rules/exclusion_ruleset.rb
```diff
@@ -95,5 +87,18 @@ def eligible_for_exclusion(is_pregnant, is_american_indian_or_alaska_native, is_
...
+    private
+
+    def meets_end_condition(member_data_exemption, certification_date)
+      return if certification_date.nil?
+      return if member_data_exemption.nil?
+
+      cert_month = certification_date.beginning_of_month
+      Array(member_data_exemption.periods).any? do |period|
+        next unless period.period_end
+        cert_month <= period.period_end.end_of_month
```
jgavin-nava: Same thing here do we need to check if the cer_month is before the period.period_start as well?

reporting-app/app/models/rules/exclusion_ruleset.rb
jgavin-nava: I dont see this constant used at all maybe an old cleanup or maybe we expect to use it later

reporting-app/app/services/exclusion_determination_service.rb
```diff
@@ -155,5 +156,13 @@ def extract_attribute(certification, attribute)
 
       certification.member_data.send(attribute)
     end
+
+    def extract_exemption(certification, attribute)
+      return nil unless certification.member_data&.exemptions.present?
+
+      certification.member_data.exemptions.select do |e|
```
jgavin-nava: Just more for my understanding and learning purposes why .select then .first versus using .find with the same block?

reporting-app/app/models/rules/exclusion_ruleset.rb
jdettmannnava: I went back and forth. Might as well, I suppose.

reporting-app/app/models/rules/exclusion_ruleset.rb
jdettmannnava: I went back and forth. Might as well, I suppose.

reporting-app/app/models/rules/exclusion_ruleset.rb
jgavin-nava: Yea maybe unnecessary but my thought was if they are supplying us with that information we can make sure both sides of the dates are compliant

reporting-app/app/services/exclusion_determination_service.rb
jdettmannnava: Nope; find is better

jgavin-nava (approved):

reporting-app/app/models/rules/exclusion_ruleset.rb
```diff
@@ -95,5 +85,18 @@ def eligible_for_exclusion(is_pregnant, is_american_indian_or_alaska_native, is_
...
+
+    private
+
+    def meets_end_condition(member_data_exemption, certification_date)
+      return if certification_date.nil?
+      return if member_data_exemption.nil?
+
+      cert_month = certification_date.beginning_of_month
+      Array(member_data_exemption.periods).any? do |period|
+        next unless period.period_start && period.period_end
```
giverm, line 97: I know this has already merged, but do we need to consider an open-ended period where the condition is ongoing? Or is the assumption that those members wouldn't be sent to OSCER in the first place?

jdettmannnava: Right now, the API tech spec has both dates filled.

PR #842, opened 2026-08-05
PBI-69963: Use new api exemptions in exception determination service

jdettmannnava:
## Ticket

Resolves PBI-69963

## Changes

`ExceptionDeterminationService` now uses the `Certifications::MemberData::Exemption` objects the new API supplies. Every check now reads a verified exemption and its periods instead of the legacy `dates_*` and boolean member-data attributes.

- **`ExceptionDeterminationService`** — most checks read exemptions through one helper that asks whether any certifiable month falls inside a period, at month granularity and with both bounds required, matching `Rules::ExclusionRuleset#meets_end_condition`. `inmate` extends the period end by `INMATE_BUFFER_MONTHS`; `former_foster_care` uses presence plus the age cap; `age_under_19` still reads `date_of_birth`.
- **New exception checks** — now that the API gives them as time-based attributes, added `veteran_disability`, `medically_frail`, and `tanf_snap_work`, all mandatory and period-based, with the reason codes `veteran_disability_excepted`, `medically_frail_excepted`, and `tanf_snap_work_excepted`.
- **`Certifications::MemberData#verified_exemption`** — shared applies-and-verified lookup; `ExclusionDeterminationService` now reads exemptions through it instead of its own copy.
- **`Rules::ExclusionRuleset#caretaker`** — now denies when a child born after the certification month.
- **API Exemption type enum** — added `inpatient_medical_care`, `declared_emergency_county`, `high_unemployment_county`, and `other_program`. Were not in API because not needed upstream, but added for consistent exception determination logic.
- **`Determination`** — `was_former_foster_care` now maps to `former_foster_care_excepted`, so every exception code follows the documented convention. All 14 exception reason codes gained English translations; they were previously rendering through the `humanize` fallback.
- **Demo create form** — the external-exception selector writes an exemption spanning the certifiable months rather than the legacy date arrays.

## Context for reviewers

The exclusion and exception checks now read the same exemptions and differ only in what they compare against — the certification month for exclusions, any certifiable month for exceptions.

Two known gaps, both out of scope here and neither introduced by this change alone:

- `UnifiedRecordProcessor` writes only legacy fields, so batch-uploaded certifications produce no exclusions (already true after #813) and now no pregnancy exception either. Every other exception was already unreachable from the batch path.
- Certifications already persisted hold the old `member_data` jsonb shape and will adjudicate differently if re-run.

## Testing

Full suite green (3331 examples, 0 failures), `make lint` reports no offenses.

Manually verified exception "Receiving inpatiant medical care" worked as expected.

♻️ Environment destroyed ♻️


reporting-app/app/models/certifications/member_data.rb
```diff
@@ -112,4 +112,13 @@ class Exemption < ValueObject
   attribute :dates_in_high_unemployment_county, :array, of: ActiveModel::Type::Date.new
   attribute :dates_traveling_for_medical_care, :array, of: ActiveModel::Type::Date.new
   attribute :dates_participating_in_other_program, :array, of: ActiveModel::Type::Date.new
```
jgavin-nava, line 114: Do we plan on bringing these back I think this is all dead code now

jgavin-nava (approved):
LGTM otherwise

reporting-app/app/models/certifications/member_data.rb
jdettmannnava: Cleanup is a separate ticket.

PR #844, opened 2026-08-06
PBI-69965: Add credit hours and enforce verification status in Cert Creation Service

jdettmannnava:
## Ticket

Resolves PBI-69965

## Changes

- `Certifications::CreationService` only creates `ExternalHourlyActivity` and `ExternalIncomeActivity` records for activities whose verification status is `verified`.
- Education activities may report `credit_hours` in place of `hours`. `Certifications::CreationService` converts them at intake using `CREDIT_HOURS_MULTIPLIER` (12.99 clock hours per credit hour, per the Federal Register).
- `Certifications::MemberData::Activity` requires `verification_status`, requires `hours` only when the activity is not an education activity reporting credit hours, and validates `credit_hours` as greater than zero.
- `Certifications::CreationService` falls back to an activity's `name` when `employer` is absent on income activities.
- OpenAPI spec: `hours` and `credit_hours` are now declared properties with an `anyOf` rule covering the either/or, `verification_status` is required and non-nullable, and `name` is documented alongside `employer`.

## Context for reviewers

`verification_status` moving from optional to required is a breaking change for any client that omits it. Without it, activities would be silently discarded and the request would still return `201`, so it fails loudly instead.

`CREDIT_HOURS_MULTIPLIER` is a `BigDecimal` rather than a float — `credit_hours` is a decimal attribute, and a float multiplier would push the result through binary floating point.

Two known gaps to address in follow-up tickets:

- **Duplicate detection.** `ExternalHourlyActivityService.duplicate_entry?` keys on `(member_id, category, hours, period_start, period_end)`. Two entries with the same hours are rejected as duplicates, rolling back the whole certification. Unlike income, the distinguishing field, `name`, is neither part of the key nor persisted for hourly activities.
- **Multi-month activity periods.** The credit-hours multiplier is monthly, so a period spanning more than one month is credited a single month. This needs handling across all activities, not just credit hours.

## Testing

Request specs cover the new API behavior end to end: an education activity reporting only `credit_hours` is accepted and stored as the converted hours, and requests are rejected when `credit_hours` appears outside the education category, when `credit_hours` is not greater than zero, and when `verification_status` is missing or invalid. Service specs cover activities being skipped when not verified, `hours` taking precedence when both values are supplied, and a non-education category with no hours failing.

```
make test    # 3340 examples, 0 failures — line 96.77%, branch 84.17%
make lint    # 569 files inspected, no offenses detected
```

Manual test with demo ("Partially met income requirement" and "Fully met work hours requirement") worked as expected.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/certifications/member_data.rb
+++ b/reporting-app/app/models/certifications/member_data.rb
@@ -30,7 +30,9 @@ class Certifications::MemberData < ValueObject
     TYPE_HOURLY = "hourly"
     TYPE_INCOME = "income"
     ACTIVITY_TYPES = [ TYPE_HOURLY, TYPE_INCOME ].freeze
-    VERIFICATION_STATUSES = %w[verified self_attested pending].freeze
+    CATEGORY_EDUCATION = "education"
+    VERIFIED = "verified"
+    VERIFICATION_STATUSES = [ VERIFIED, "self_attested", "pending" ].freeze
 
     attribute :type, :string
     attribute :category, :string
@@ -47,7 +49,8 @@ class Certifications::MemberData < ValueObject
 
     validates :type, presence: true, inclusion: { in: ACTIVITY_TYPES }
     validates :category, presence: true, inclusion: { in: ::Activity::ALLOWED_CATEGORIES }
-    validates :hours, presence: true, if: -> { type == TYPE_HOURLY }
+    validates :hours, presence: true, if: -> { type == TYPE_HOURLY && !education_credit_hours? }
+    validates :credit_hours, numericality: { greater_than: 0 }, allow_nil: true
     validates :gross_income, presence: true,
                              numericality: { greater_than: 0 },
                              if: -> { type == TYPE_INCOME }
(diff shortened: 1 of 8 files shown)


reporting-app/app/models/certifications/member_data.rb
```diff
@@ -47,7 +53,9 @@ class Activity < ValueObject
 
     validates :type, presence: true, inclusion: { in: ACTIVITY_TYPES }
     validates :category, presence: true, inclusion: { in: ::Activity::ALLOWED_CATEGORIES }
-    validates :hours, presence: true, if: -> { type == TYPE_HOURLY }
+    validates :hours, presence: true, if: -> { type == TYPE_HOURLY && !education_credit_hours? }
+    validates :credit_hours, numericality: { greater_than: 0 }, allow_nil: true
+    validates :credit_hours, absence: true, unless: -> { category == CATEGORY_EDUCATION }
```
jgavin-nava: Should we make sure the the type is hourly as well for this? 


```suggestion
    validates :credit_hours, absence: true, unless: -> { type == TYPE_HOURLY && category == CATEGORY_EDUCATION }
```

jgavin-nava (approved):
LGTM

PR #855, opened 2026-08-11
PBI-69962: Prevent duplicate certification requests via the API

jgavin-nava:
## Ticket

PBI-69962 — Prevent duplicate certification requests from being created in OSCER via the API

## Changes

- Persist `application_date` from `POST /api/certifications` via `Api::Certifications::CreateRequest`
- Add `Certification.find_duplicate` keyed on `member_id` + `case_number` + `application_date`
- In `Api::CertificationsController#create`, return the existing certification as `200` with `Api::Certifications::Response` instead of creating another row
- Skip the duplicate check when `member_id`, `case_number`, or `application_date` is blank (blank dates are not treated as matches)
- Document the `200` duplicate response in the generated OpenAPI spec
- Request/model specs for replay, outcome status on the existing cert, `respond-async`, differing date, and omitted date

## Context for reviewers

- App-level lookup only — no unique DB index on this branch. Concurrent overlapping POSTs can still both insert.
- `application_date` is not required on create yet. The blank guard is in place so missing dates never match; requiring the field can be a follow-up.
- OpenAPI still documents `application_date` as optional/nullable.
- Unique-index work (if any) lives separately and is out of scope here.

## Testing

- From `reporting-app/`: `make test args='spec/requests/api/certifications_spec.rb spec/models/certification_spec.rb'`
- 60 examples passed (SimpleCov `exit 2` is expected on a partial suite)

## Local testing

Setup
```
export API_SECRET_KEY='your-local-secret'   # same value the app has

BODY='{
  "member_id": "dup-curl-1",
  "case_number": "C-DUP-1",
  "application_date": "2025-10-16",
  "member_data": {
    "name": { "first": "John", "last": "Doe" }
  },
  "certification_requirements": {
    "certification_date": "2025-10-16",
    "number_of_months_to_certify": 1,
    "lookback_period": 3,
    "due_period_days": 30
  }
}'

post_cert() {
  local payload="$1"
  local extra_header="${2:-}"
  SIG=$(printf '%s' "$payload" | openssl dgst -sha256 -hmac "$API_SECRET_KEY" -binary | base64)
  curl -sS -D /tmp/cert-headers -o /tmp/cert-body \
    -X POST http://localhost:3000/api/certifications \
    -H "Content-Type: application/json" \
    -H "Authorization: HMAC sig=$SIG" \
    ${extra_header:+-H "$extra_header"} \
    -d "$payload"
  echo "HTTP $(awk 'NR==1{print $2}' /tmp/cert-headers)"
  cat /tmp/cert-body | python3 -m json.tool
}
```

Calls and returns
```
post_cert "$BODY"          # 201, note id
post_cert "$BODY"          # 200, same id, no new row
```

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/controllers/api/certifications_controller.rb
+++ b/reporting-app/app/controllers/api/certifications_controller.rb
@@ -36,6 +36,7 @@ class Api::CertificationsController < ApiController
   # @request_body_example Without explicit months and due date [Reference:#/components/schemas/CertificationCreateRequestBody/examples/without_explicit_months_that_can_be_certified_and_due_date]
   # @request_body_example Certification type [Reference:#/components/schemas/CertificationCreateRequestBody/examples/certification_type]
   # @response Created Certification.(201) [Reference:#/components/schemas/CertificationResponseBody]
+  # @response Existing Certification returned for a duplicate request.(200) [Reference:#/components/schemas/CertificationResponseBody]
   # @response Accepted(202) []
   # @response User error.(400) [Reference:#/components/schemas/ErrorResponseBody]
   # @response User error.(422) [Reference:#/components/schemas/ErrorResponseBody]
@@ -49,6 +50,17 @@ class Api::CertificationsController < ApiController
     # Build certification from request
     certification = create_request.to_certification
     authorize certification
+
+    # Idempotency: if a certification already exists for this duplicate key,
+    # return its status instead of creating another row.
+    existing = Certification.find_duplicate(
+      member_id: certification.member_id,
+      case_number: certification.case_number,
+      application_date: certification.application_date
+    )
+    if existing
+      return render_data(Api::Certifications::Response.from_certification(existing))
+    end
 
     begin
       service = Certifications::CreationService.new(certification)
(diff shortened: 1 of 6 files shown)


reporting-app/app/controllers/api/certifications_controller.rb
```diff
@@ -50,6 +51,17 @@ def create
...
 
+    # Idempotency: if a certification already exists for this duplicate key,
+    # return its status instead of creating another row.
+    existing = Certification.find_duplicate(
+      member_id: certification.member_id,
+      case_number: certification.case_number,
+      application_date: certification.application_date
+    )
+    if existing
+      return render_data(Api::Certifications::Response.from_certification(existing))
```
giverm, line 62: When there is no existing `certification`, the response has a different shape depending on the value of `request.headers["prefer"]`. Do we need to follow the same pattern here?

giverm (commented):
I had one question about the response, but the rest looks good to me

reporting-app/app/controllers/api/certifications_controller.rb
jgavin-nava: I am not sure if that shape would be needed in this situation. We would have cert already created so the status wouldnt be pending it would have data behind it so they wouldnt be waiting for processing the creation of that cert.

I think responding the same way no matter that header if there is a duplicate would give us the same result and same thing they would want in that situation. From my understanding they would send us that header because they did not want to wait for the creation of that cert and the response might take a little. Since it is already there they wouldnt need to wait. 

Sorry I think I said the same thing twice hahah.

giverm (approved):

PR #856, opened 2026-08-11
PBI-69965: Add household income to external income

jdettmannnava:
## Ticket

Resolves PBI-69965

## Changes

- `Certifications::HouseholdData` now carries per-period gross income for each household member (`GrossIncome`), replacing the single `monthly_income_dollars_total` field, with validations on the amount and the period.
- `Certifications::CreationService` creates an `ExternalIncomeActivity` for each household member's reported gross income, skipping the applicant when they also appear in the household.
- Tax IDs on `Certifications::HouseholdData::Member` and `Certifications::MemberData` use the Strata `:tax_id` attribute, so SSNs arriving as JSON strings are normalized and format-validated.
- `Api::Certifications::CreateRequest` and `Certification` accept and persist `household_data`; the OpenAPI spec is regenerated.

## Context for reviewers

Household income counts toward the member's income compliance — rows are written against `certification.member_id`, so `IncomeComplianceDeterminationService` and `MemberDashboardCompliance` pick them up like any other external income.

Applicant matching lives in `Certifications::HouseholdData::Member#same_person_as?`: tax ID when both sides have one, otherwise first and last name plus date of birth must all match. An unconfirmable match returns false, so income is never silently dropped.

`MemberData#ssn` previously used `ActiveModel::Type::Json`, whose cast only handles Hashes — a JSON string SSN became `nil`, which meant the value was silently discarded on every API-created certification. Worth flagging as a behavior change: a malformed SSN in either `member_data` or `household_data` now returns a 422 where it was previously accepted and dropped.

- the household income `category` is hardcoded to `household`, which can be duplicated
- rows carry no attribution to the household member they came from, so they are indistinguishable from the applicant's own income in the member and staff views

## Testing

`make lint` clean, `make test` green — 3376 examples, 0 failures (96.79% line / 84.39% branch).

- New `Certifications::HouseholdData` model spec covers the gross income validations, tax ID casting and normalization, and every `#same_person_as?` branch.
- `Certifications::CreationService` spec asserts one `ExternalIncomeActivity` per household gross income by amount, and that the applicant is skipped whether listed with an exact tax ID, a dash-formatted one, or none at all.
- API request specs cover household income creating income rows and no hourly rows, plus 422s for a gross income missing its period and for a malformed SSN.

♻️ Environment destroyed ♻️


jgavin-nava:
I am little confused on the call out for this? 

> two household members reporting the same amount for the same period collide with ExternalIncomeActivityService's duplicate check and roll back the whole certification

Its seems like without fixing this now we are adding a bug and what the discussion is that needs to be had for it?

jdettmannnava:
> I am little confused on the call out for this?
> 
> > two household members reporting the same amount for the same period collide with ExternalIncomeActivityService's duplicate check and roll back the whole certification
> 
> Its seems like without fixing this now we are adding a bug and what the discussion is that needs to be had for it?

We have some open questions on category, reporting, whether we will be using "gross" here, etc. I have a sense that anything I do to dedupe here is going to be overwritten by future decisions.

jgavin-nava (approved):
LGTM

reporting-app/app/models/external_income_activity.rb
```diff
@@ -14,7 +14,7 @@
 class ExternalIncomeActivity < ApplicationRecord
   include Strata::Attributes
 
-  ALLOWED_CATEGORIES = ActivityCategories::ALL
+  ALLOWED_CATEGORIES = ActivityCategories::ALL + %w[household]
```
jgavin-nava: Instead of hardcoding household can we make a constant and then we can use it later as well. 


```suggestion
  CATEGORY_HOUSEHOLD = "household"
  ALLOWED_CATEGORIES = (ActivityCategories::ALL + [ CATEGORY_HOUSEHOLD ]).freeze
```

reporting-app/app/services/certifications/creation_service.rb
```diff
@@ -82,4 +86,26 @@ def create_external_income_activities
...
+  def create_household_income_activities
+    return unless certification.household_data&.members.present?
+
+    certification.household_data.members.each do |household_member|
+      next if household_member.same_person_as?(certification.member_data)
+
+      Array(household_member.gross_incomes).each do |gross_income|
+        ExternalIncomeActivityService.create_entry(
+          member_id: certification.member_id,
+          category: "household",
```
jgavin-nava:
```suggestion
          category: ExternalIncomeActivity::CATEGORY_HOUSEHOLD
```

reporting-app/app/services/external_income_activity_service.rb
```diff
@@ -77,6 +77,8 @@ def maybe_recalculate_income_compliance(member_id)
 
     # Same dimensions as ExternalHourlyActivityService duplicate check; source_type is not part of the key.
     def duplicate_entry?(member_id:, category:, gross_income:, period_start:, period_end:)
+      return if category == "household"
```
jgavin-nava:
```suggestion
      return if category == ExternalIncomeActivity::CATEGORY_HOUSEHOLD
```

jgavin-nava (approved):
LGTM otherwise

reporting-app/app/services/certifications/creation_service.rb
```diff
@@ -82,4 +86,26 @@ def create_external_income_activities
...
+  def create_household_income_activities
+    return unless certification.household_data&.members.present?
+
+    certification.household_data.members.each do |household_member|
+      next if household_member.same_person_as?(certification.member_data)
+
+      Array(household_member.gross_incomes).each do |gross_income|
+        ExternalIncomeActivityService.create_entry(
+          member_id: certification.member_id,
+          category: ExternalIncomeActivity::CATEGORY_HOUSEHOLD
```
jgavin-nava:
```suggestion
          category: ExternalIncomeActivity::CATEGORY_HOUSEHOLD,
```

jgavin-nava (approved):

PR #857, opened 2026-08-14
Add education status to api education activities

jdettmannnava:
## Ticket

Resolves PBI-69965

## Changes

- `Certifications::HouseholdData` now carries per-period gross income for each household member (`GrossIncome`), replacing the single `monthly_income_dollars_total` field, with validations on the amount and the period.
- `Certifications::CreationService` creates an `ExternalIncomeActivity` for each household member's reported gross income, skipping the applicant when they also appear in the household.
- Tax IDs on `Certifications::HouseholdData::Member` and `Certifications::MemberData` use the Strata `:tax_id` attribute, so SSNs arriving as JSON strings are normalized and format-validated.
- `Api::Certifications::CreateRequest` and `Certification` accept and persist `household_data`; the OpenAPI spec is regenerated.

## Context for reviewers

Household income counts toward the member's income compliance — rows are written against `certification.member_id`, so `IncomeComplianceDeterminationService` and `MemberDashboardCompliance` pick them up like any other external income.

Applicant matching lives in `Certifications::HouseholdData::Member#same_person_as?`: tax ID when both sides have one, otherwise first and last name plus date of birth must all match. An unconfirmable match returns false, so income is never silently dropped.

`MemberData#ssn` previously used `ActiveModel::Type::Json`, whose cast only handles Hashes — a JSON string SSN became `nil`, which meant the value was silently discarded on every API-created certification. Worth flagging as a behavior change: a malformed SSN in either `member_data` or `household_data` now returns a 422 where it was previously accepted and dropped.

Known and deliberately deferred to a separate discussion (see the FIXME in `create_household_income_activities`):

- the household income `category` is hardcoded to `employment`
- rows carry no attribution to the household member they came from, so they are indistinguishable from the applicant's own income in the member and staff views
- two household members reporting the same amount for the same period collide with `ExternalIncomeActivityService`'s duplicate check and roll back the whole certification

## Testing

`make lint` clean, `make test` green — 3376 examples, 0 failures (96.79% line / 84.39% branch).

- New `Certifications::HouseholdData` model spec covers the gross income validations, tax ID casting and normalization, and every `#same_person_as?` branch.
- `Certifications::CreationService` spec asserts one `ExternalIncomeActivity` per household gross income by amount, and that the applicant is skipped whether listed with an exact tax ID, a dash-formatted one, or none at all.
- API request specs cover household income creating income rows and no hourly rows, plus 422s for a gross income missing its period and for a malformed SSN.

Manually tested, and works as expected (including recording the enrollment status in the decision).

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/certifications/member_data.rb
+++ b/reporting-app/app/models/certifications/member_data.rb
@@ -34,6 +34,11 @@ class Certifications::MemberData < ValueObject
     VERIFIED = "verified"
     VERIFICATION_STATUSES = [ VERIFIED, "self_attested", "pending" ].freeze
 
+    ENROLLMENT_FULL_TIME = "full_time"
+    ENROLLMENT_HALF_TIME = "half_time"
+    ENROLLMENT_STATUSES = [ ENROLLMENT_FULL_TIME, ENROLLMENT_HALF_TIME, "less_than_half_time" ].freeze
+    QUALIFYING_ENROLLMENT_STATUSES = [ ENROLLMENT_FULL_TIME, ENROLLMENT_HALF_TIME ].freeze
+
     # One credit hour is 12.99 clock hours per month (3 hours per week over a 4.33 week month)
     # per the Federal Register.
     CREDIT_HOURS_MULTIPLIER = BigDecimal("12.99")
@@ -50,10 +55,12 @@ class Certifications::MemberData < ValueObject
     attribute :employer, :string
     attribute :name, :string
     attribute :verification_status, :string
+    attribute :enrollment_status, :string
 
     validates :type, presence: true, inclusion: { in: ACTIVITY_TYPES }
     validates :category, presence: true, inclusion: { in: ::Activity::ALLOWED_CATEGORIES }
-    validates :hours, presence: true, if: -> { type == TYPE_HOURLY && !education_credit_hours? }
+    validates :hours, presence: true,
+                      if: -> { type == TYPE_HOURLY && !education_credit_hours? && !education_enrollment? }
     validates :credit_hours, numericality: { greater_than: 0 }, allow_nil: true
     validates :credit_hours, absence: true, unless: -> { type == TYPE_HOURLY && category == CATEGORY_EDUCATION }
     validates :gross_income, presence: true,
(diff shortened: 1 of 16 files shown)


reporting-app/app/services/community_engagement_check_service.rb
```diff
@@ -55,7 +55,10 @@ def assess(certification)
       Assessment.new(
         hours_data: hours_data,
         income_data: income_data,
-        hours_ok: HoursComplianceDeterminationService.compliant_for_total_hours?(hours_data[:total_hours]),
+        # A qualifying education enrollment meets the hours requirement outright, so it passes the
+        # hours track instead of standing up a third one the determination payload has no shape for.
+        hours_ok: HoursComplianceDeterminationService.compliant_for_total_hours?(hours_data[:total_hours]) ||
+          HoursComplianceDeterminationService.education_enrollment_compliant?(certification),
```
jgavin-nava, line 61: Just want to call out that this could cause issues with auditing why a member was compliant looking like it has 0 total hours but still being compliant. If product is ok with that then fine with me, if not maybe we can just add a key to the hours data like `:eductation_enrollment_compliant` just to track that is how they were compliant.

reporting-app/app/models/certifications/member_data.rb
```diff
@@ -34,6 +34,11 @@ class Activity < ValueObject
     VERIFIED = "verified"
     VERIFICATION_STATUSES = [ VERIFIED, "self_attested", "pending" ].freeze
 
+    ENROLLMENT_FULL_TIME = "full_time"
+    ENROLLMENT_HALF_TIME = "half_time"
+    ENROLLMENT_STATUSES = [ ENROLLMENT_FULL_TIME, ENROLLMENT_HALF_TIME, "less_than_half_time" ].freeze
```
jgavin-nava, line 40: Nit-pick: but can we just make this consistent

```suggestion
    ENROLLMENT_LESS_THAN_HALF_TIME = "less_than_half_time"
    ENROLLMENT_STATUSES = [ ENROLLMENT_FULL_TIME, ENROLLMENT_HALF_TIME, ENROLLMENT_LESS_THAN_HALF_TIME ].freeze
```

jdettmannnava: Claude did this, and I actually like it as is, as it indicates the value is not used anywhere else. But I will change it on any pushback!

reporting-app/app/services/community_engagement_check_service.rb
jdettmannnava: I added the enrollment status to the blob that ends up in the determination data.

reporting-app/app/models/certifications/member_data.rb
jgavin-nava: That is fine with me. I did notice it wasnt anywhere else so that makes sense.

jgavin-nava (approved):
LGTM

PR #866, opened 2026-08-24
PBI 70715: Chunk multi-month activities into months.

jdettmannnava:
## Ticket

Resolves PBI-70715

## Changes

- `create_entries` added to external activity services; called by certifications creation service
- apportionment utilities added to ActivityAggregator

## Context for reviewers

This breaks activities that span more than one month into chunks that lie within a single month. All hourly activities, and income activities that do not include only whole months are apportioned by days. Income activities that cover entire months are presumed to be split by month rather by the number of days in the month.

Errors are generated by passing dubious data down to the `create_entry` method to maintain consistency.

## Testing

Extensive CI tests.

Manual hits against the API worked as expected
- Hourly activity spanning 2 months
- Income activity spanning 2 months (not start on first of month)
- Income activity spanning 2 entire months

♻️ Environment destroyed ♻️

### DIFF
--- a/.github/workflows/ci-infra.yml
+++ b/.github/workflows/ci-infra.yml
@@ -84,7 +84,7 @@ jobs:
         # There is currently no way to specify the checkov version to pin to (See https://github.com/bridgecrewio/checkov-action/issues/41)
         # so we need to pin the version of the checkov-action, which indirectly pins the checkov version.
         # In this case, checkov-action v12.3107.0 is mapped to checkov v3.3.1.
-        uses: bridgecrewio/checkov-action@v12.3117.0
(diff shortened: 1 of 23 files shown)


giverm (approved):
This looks great! 

One thing I noticed when reviewing this PR that's probably worth pointing out; The duplication check doesn't take the employer into account. So, if someone is working part time for 2 employers and makes the same amount/works the same number of hours  in the same period, the second entry would be rejected. I bring it up here, because I could imagine splitting the periods into months increases the likelihood of activities being incorrectly marked as duplicate, rolling back the whole certification.

Happy to talk through this with you if I'm not making sense.

PR #873, opened 2026-08-25
Wire certification period bounds through certifications API

giverm:
## Changes

- Add `certification_period_start` and `certification_period_end` attributes to `Certifications::RequirementParams`, and pass both through `to_requirements`.
- Document both fields in `lib/assets/oas.json` as optional, nullable dates on all three request shapes that already carry `certification_date`, and add them to two request examples. Regenerated `openapi.generated.yml` to match.
- Add `spec/models/certifications/requirements_spec.rb`, covering the bounds through the full-requirements request shape, the parameter shape, batch upload input, and a database round trip.
- Add `RequirementParams#to_requirements` specs for both supplied and omitted bounds.
- Add request specs asserting that supplied bounds are stored and that omitted bounds stay null.

## Context for reviewers

Both attributes were already declared on `Certifications::Requirements`, but nothing else in the application or its specs referenced them. `RequirementParams#to_requirements` did not pass them through, so they could not be set from a parameter-shaped request, and they were missing from the published API spec, so a caller had no way to send them at all. They were effectively unreachable. This makes them settable end to end.

**The bounds are supplied, never derived.** On a recertification or a change in circumstance they describe the member's currently active, expiring certification period, which is the window the caller is renewing out of. A new application has no certification period at all, because the member has no prior coverage to describe, so its month anchor is the application date instead. That is why there is no derivation and no presence validation here: when a request omits the bounds, they stay null, and a spec asserts that on the model, through `to_requirements`, and through a live API request.

This is deliberately additive and safe to land on its own, because nothing reads either field before or after the change. Repointing the exclusion rules and the member dashboard onto the new anchor is separate follow-up work, and `certification_date` is untouched here.

One fixture convention, called out because it is easy to copy wrongly: the examples use a certification period that *precedes* the request (certification date 2026-05-20 inside a 2026-01-01 to 2026-06-30 period) rather than one starting after it. The period being described is the one currently running out, not an upcoming one, and the spec comments say so at each site.

## Testing

```
make lint   # 584 files inspected, no offenses detected
make test   # 3469 examples, 0 failures
```

Line coverage 96.83%, branch coverage 84.25%, both above the configured minimums. No determination outcomes change, which the existing exclusion and determination specs confirm by passing without expectation edits.

Targeted re-run of the three affected spec files after the final fixture change: 62 examples, 0 failures.

No manual or end-to-end verification was done. Nothing reads these fields yet, so there is no user-visible behaviour to exercise.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/spec/models/certifications/requirements_spec.rb
+++ b/reporting-app/spec/models/certifications/requirements_spec.rb
@@ -78,7 +78,7 @@ RSpec.describe Certifications::Requirements do
       requirements = CertificationService.new.certification_requirements_from_input(
         bounds.merge(
           "certification_date" => "2026-05-20",
-          "certification_type" => "new_application"
+          "certification_type" => "recertification"
         )
       )
 
--- a/reporting-app/spec/models/certifications/requirements_spec.rb
+++ b/reporting-app/spec/models/certifications/requirements_spec.rb
@@ -40,7 +40,7 @@ RSpec.describe Certifications::Requirements do
     end
   end
 
-  describe "the bounds survive the request and storage paths" do
+  describe "supplied period bounds" do
     let(:bounds) do
       {
         "certification_period_start" => "2026-01-01",
@@ -48,7 +48,7 @@ RSpec.describe Certifications::Requirements do
       }
     end
 
-    it "are carried when the request supplies full requirements as JSON" do
+    it "survives a full-requirements request" do
       input = Api::Certifications::RequirementsOrParamsInput.new(
         bounds.merge(
           "certification_date" => "2026-05-20",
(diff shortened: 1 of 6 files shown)


reporting-app/spec/models/certifications/requirements_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+
+      expect(input).to be_a(Certifications::RequirementParams)
+      expect(input.to_requirements.certification_period_start).to eq Date.new(2026, 1, 1)
+    end
+
+    it "survives batch upload input" do
+      requirements = CertificationService.new.certification_requirements_from_input(
+        bounds.merge(
+          "certification_date" => "2026-05-20",
+          "certification_type" => "new_application"
```
jgavin-nava: From the description of this spec it states that.

>   The certification period is supplied by the caller and never derived here. On a
   recertification it is the member's currently active, expiring period; a new
   application has none, because the member has no prior coverage period.

should this be or am I understanding it wrong?
```suggestion
          "certification_type" => "recertification"
```

jgavin-nava (approved):
One small nit-pick: LGTM otherwise

reporting-app/spec/models/certifications/requirements_spec.rb
giverm: Good catch! Thank you

PR #875, opened 2026-08-27
PBI-70714: validate by month instead of total

jdettmannnava:
## Ticket

Resolves PBI-70714

## Changes

- `HoursComplianceDeterminationService` and `IncomeComplianceDeterminationService` now decide compliance from per-month totals instead of the lookback aggregate: `compliant_for_monthly_hours?` / `compliant_for_monthly_income?` replace the former `compliant_for_total_*` checks and pass when any single month reaches the threshold.
- Both aggregates gained an `hours_by_month` / `income_by_month` map, built in `ActivityAggregator` and merged across external and member-reported sources by the shared `merge_external_with_member_data`.
- `ExternalHourlyActivity` and `ExternalIncomeActivity` expose `#month` so external rows key by month alongside member `Activity` rows.
- `CommunityEngagementCheckService` assesses both tracks against the monthly maps, so the combined external CE step and `DataSourceCheckService` follow the same rule.
- `Determinations::HoursBasedDeterminationData` and `Determinations::IncomeBasedDeterminationData` serialize a `maximum_monthly_hours` / `maximum_monthly_income` attribute — the best single month, which is the figure the outcome actually turns on — so a stored determination explains itself rather than showing only a lookback total that contradicts it.
- `ActivityAggregator` summaries load their rows once and total them in memory rather than mixing an in-memory `by_month` pass with SQL aggregates; the `.reorder(nil)` workaround in `HoursComplianceDeterminationService` went with the SQL path.

## Context for reviewers

The threshold constants were already monthly (`TARGET_HOURS`, `TARGET_INCOME_MONTHLY`) — only the comparison changes. Multi-month intake is chunked per month upstream (PBI-70715), so each external row falls in one month and `#month` can key off `period_start`.

`maximum_monthly_*` is optional rather than required: it is nil when an aggregate omits the monthly map, and serializes as nil rather than a misleading zero. Every production path supplies the map, so the field is populated on real determinations.

Summarize methods no longer differentiate ActiveRecord::Relations and rows, as we need to load everything to get the monthly aggregations, so the ActiveRecord::Releation optimization no longer held.

One thing worth a decision, left out of this PR: the member dashboard, the insufficiency emails, and the staff case hours/income tables still compute "needed" as `target - total`. Under the monthly rule a member with 40 hours in each of two months is `not_compliant` while those surfaces report the requirement met. Same shape for income.

## Testing

`make lint` clean (574 files, no offenses) and `make test` green — 3521 examples, 0 failures, line coverage 96.85%, branch 84.09%.

♻️ Environment destroyed ♻️


reporting-app/app/services/member_dashboard_compliance_service.rb
```diff
@@ -19,9 +19,8 @@ def build(certification:, certification_case:, activity_report_application_form:
       show_income = income_summary_visible?(latest, member_status) && lookback.present?
 
       external_hourly_rel = lookback.present? ? aggregator.fetch_external_hourly_activities(certification) : ExternalHourlyActivity.none
-      # Pass only the external relation: +HoursComplianceDeterminationService+ will fetch and
-      # +.reorder(nil)+ member rows internally so its +GROUP BY :category+ aggregation hits
-      # the Postgres SUM path instead of erroring on the +ORDER BY+ on +member_hour_activities_for_certification+.
+      # Pass only the external relation, which the dashboard tables need anyway;
+      # +HoursComplianceDeterminationService+ fetches the member rows itself.
```
jgavin-nava, line 23: is the reason you arent making changes to the member dashboard because we arent going to most likely use this? So its not worth the changes? There are a few spots in here that still would use the old way

jgavin-nava (approved):
If its not worth making the changes to the member dashboard then the rest LGTM

reporting-app/app/services/member_dashboard_compliance_service.rb
jdettmannnava: This is more the "we are going to be redesigning the whole thing so don't fudge things now".

PR #877, opened 2026-08-28
PBI-70896: Record determination time instead of the certification date

giverm:
## Ticket

Resolves PBI-70896

## Changes

- All eight `record_determination!` calls in `CertificationCase` pass
`determined_at: Time.current` instead of
`certification.certification_requirements.certification_date`.
- The three `determined_at` writes in `Dev::MemberDashboardFrameSetup`
do the same.
- The six determination specs that asserted `determined_at` was
`be_present` now assert exact equality under a frozen clock, with one
`around` hook per describe.

## Context for reviewers

Every determination on a certification carried an identical timestamp at
midnight, describing the period under review rather than the moment a
decision was made. The three readers already interpret the column as the
time a decision happened, so none changed, and no backfill is needed since
the API has no live consumers.

`Dev::MemberDashboardFrameSetup` has no spec coverage, so those three
lines were verified by running the frames against a development database.

Also removes the last `certification_date` reference from both files.

## Testing

Reintroducing the defect at all eight call sites fails 6 of the 57
examples in the file, so the specs discriminate rather than merely pass.

```
make lint   # 584 files inspected, no offenses detected
make test   # 3504 examples, 0 failures
```

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/certification_case.rb
+++ b/reporting-app/app/models/certification_case.rb
@@ -67,7 +67,7 @@ class CertificationCase < Strata::Case
         reasons: [ Determination::REASON_CODE_MAPPING[:hours_reported_compliant] ],
         outcome: :compliant,
         determination_data: Determinations::HoursBasedDeterminationData.from_aggregate(hours_data).to_h,
-        determined_at: certification.certification_requirements.certification_date,
+        determined_at: Time.current,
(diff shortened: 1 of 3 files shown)


jdettmannnava (approved):
LGTM

PR #879, opened 2026-08-31
PBI-70821 add name to duplicate check

jdettmannnava:
## Ticket

Resolves PBI-70821

## Changes

- Added `name` and `origin_hash` (indexed) columns to `external_hourly_activities` and `external_income_activities`.
- Added `OriginHash`, a concern that digests the values identifying one external activity submission, normalizing them first so decimal scale, numeric or date class, and name casing/padding don't change the result.
- `ExternalHourlyActivityService` and `ExternalIncomeActivityService` now persist the reported `name` on every entry, and fingerprint the submission in `create_entries` — before it is split into monthly entries — rather than duplicate-checking each persisted row.
- A submission already on file is logged and skipped: `create_entries` returns no entries, writes nothing, and lets the caller carry on with the rest of its work.
- `Certifications::CreationService` passes the reported name through on both the hourly and income calls, and identifies household income by the household member's name, tax ID, and date of birth.

## Context for reviewers

The duplicate check used to run per persisted row, comparing member, category, amount, and period. Two genuinely different activities that happened to share those five values — the same hours at two different employers in the same month — collided, and the second one failed the whole certification. Folding the name into the identity fixes that, and `name` is now a column rather than a value that only existed in transit.

Checking in `create_entries` rather than `create_entry` follows from where the identity lives: a submission is one reported activity, and the monthly rows are an implementation detail of storing it. All rows from one call share a hash, so a resubmission is skipped whole rather than partially applied. The tradeoff is that only an exact repeat is caught now — a resubmission overlapping a previous period no longer collides month by month.

Four things worth a look:

- A duplicate is no longer an error. It was raising `ActiveRecord::RecordInvalid`, which rolled back the certification and everything created alongside it; now it is a logged warning and the certification is created with the duplicate omitted. Callers that counted on the raise are gone.
- Following from that, a client that posts a certification containing a repeated activity now gets a 201 with the row quietly omitted, where it used to get a 422 naming the duplicate. `Api::Certifications::Response` doesn't enumerate created activities, so the omission is visible only in our logs, and the OAS annotations on `Api::CertificationsController#create` still advertise 422 for user error. Worth deciding whether integrators need a signal here.
- The duplicate warning names the member, category, period, and hash, but not the reported name — for household income that name belongs to a household member, and logs carry no PII. The hash already tells two submissions apart.
- Income identity is the reported `name`; `employer` continues into `metadata` but is no longer part of the fingerprint, so two income submissions differing only by employer are now the same submission. The API documents `name` as the field for the reporting organization, and `Certifications::CreationService` still falls back to the name for the employer metadata.

Household income is fingerprinted like everything else now, so a household member's income is recorded once. Because two household members can legitimately report the same amount for the same period, the fingerprint takes the member's name plus the tax ID and date of birth that `Certifications::HouseholdData::Member#same_person_as?` already treats as identity — two members who share a name stay distinct. Those rows now carry the reporting member's name, which incidentally gives them attribution they previously lacked.

That check is member-scoped rather than certification-scoped, so it also spans certifications: when a member recertifies and the same household income is reported again, no new rows are written. Compliance is unaffected — the original rows are still in the member's lookback window — but `reported_at` and `created_at` stay at the first report, so a re-report never refreshes recency. This already applied to member income and hourly activities; household income is newly subject to it.

`name` is nullable and unvalidated, but a blank name still participates in the fingerprint (`nil`, `""`, and `"  "` normalize alike), so an unnamed resubmission is caught.

Rows written before this migration carry a NULL hash; `duplicate_entry?` short-circuits on blank, so they never match anything. No backfill — the reported name isn't persisted on historical rows, so their hashes can't be reconstructed.

Nothing displays the stored name yet: the member dashboard and the staff tables still label external rows `"<Category> Activity"`. Left as a follow-up so both change together, and the comment in `MemberDashboardCompliance` that explained the label as missing data is updated to say the data now exists.

## Testing

`make lint` clean (577 files, no offenses) and `make test` green (3549 examples, 0 failures; line 96.86%, branch 84.00%).

Migrations verified by round trip — `make db-rollback` drops the columns, `make db-migrate` reproduces an identical `schema.rb`.

New and updated coverage in the service specs:

- The same submission twice creates no rows, returns no entries, and logs the skip; the same activity under a different name, or for a different member, is accepted.
- Income recalculates compliance only for the submission that created entries; the test was checked against a mutated service that recalculates on the duplicate path, and fails there.
- Employer is not identity: a resubmission under a different employer is still a duplicate.
- All entries from one call share a hash and carry the reported name; the hash changes when the name changes.
- Hash normalization: name casing, surrounding whitespace, and a blank-vs-absent name are ignored, and hours compare equal across Integer, Float, and BigDecimal (the API sends an integer; the chunking specs derive hours from a date range, which is a Rational).
- Household income: a repeated report from the same member is skipped, while another member's identical income, a shared name with a different tax ID and date of birth, a different amount, and a different period are all recorded.
- `Certifications::CreationService`: two hourly activities differing only by name create two rows with distinct hashes; a duplicate hourly or income activity in one request is skipped and the certification is still created; each household entry is attributed to the member who reported it.
- `POST /api/certifications` returns 201 with the duplicate income activity omitted, where it previously returned 422.

♻️ Environment destroyed ♻️


reporting-app/app/services/certifications/creation_service.rb
```diff
@@ -59,7 +59,8 @@ def create_external_hourly_activities
         period_start: activity_data.period_start,
         period_end: activity_data.period_end,
         source_type: ExternalHourlyActivity::SOURCE_TYPES[:api],
-        source_id: nil
+        source_id: nil,
+        name: activity_data.name
```
jgavin-nava, line 63: I know it was stated in the PR description but wanted to bring it up still just to get clarity. We dont want to also use emplyer here to check for a duplicate since the employmenty payloads use employer and not name?

jgavin-nava (approved):
LGTM otherwise. I think you answered it already just wanted clarity

giverm (approved):

reporting-app/app/services/certifications/creation_service.rb
jdettmannnava: In the tech spec for AK, there is only name

PR #880, opened 2026-08-31
PBI-70930: Accept the supplied due date instead of deriving it

giverm:
## Ticket

Resolves PBI-70930

## Changes

- Keep a caller-supplied `due_date` instead of discarding it in `set_type_params`
- Anchor the derived due date on the processing date (`Date.current`) rather than `certification_date`
- Add `DEFAULT_DUE_PERIOD_DAYS` to `Certifications::RequirementTypeParams` and drop the three identical per-type literals
- Move `set_due_date_from_period` from `after_validation` to `before_validation`, and add `validates :due_date, presence: true`
- Remove the one-or-the-other `due_date` / `due_period_days` validation pair
- Drop `due_period_days` from the `CertificationRequirementParameters` required list in `oas.json` and the generated spec
- Declare `due_date` on `CertificationRequirementParameters`, which honors the field but never documented it
- Reject a `due_date` that is not a date, rather than persisting it

## Context for reviewers

ARIES supplies a due date and OSCER was discarding it, deriving its own from `certification_date`. This is one of a series removing `certification_date` from the certifications API and model (ADO 70259), following PBI 70717 (#873, period bounds) and PBI 70896 (#877, determination time); the field's uses split four ways and the removal itself is gated on each of them landing.

Two things to note: the fallback carries its own default (`due_period_days || DEFAULT_DUE_PERIOD_DAYS`) rather than trusting the attribute default, because an explicit JSON `null` overrides an ActiveModel `default:`; without it, `"due_period_days": null` with no due date would persist a nil deadline and crash `MemberMailer` after the row committed. That request was a 422 before this change. And the derivation moved to `before_validation` so `validates :due_date, presence: true` has something to guard, since an `after_validation` callback produces a value nothing can check.

Keeping a caller-supplied date meant it needed a type check. ActiveModel's date cast passes non-String values through untouched, so an array or integer would validate and land in the `certification_requirements` jsonb, and `&.due_date&.strftime` does not help because `&.` guards nil rather than the wrong class. The readers then fail permanently for that record, including `MemberMailer` inside the notification chain that runs within the create request. The wipe this PR removes was incidentally the only thing preventing that.

The `oas.json` edit is here because this change caused it: removing the validation pair left the published spec marking `due_period_days` required while OSCER no longer does. The generated YAML carries one hand-applied line because `make openapi-spec` rewrites random component identifiers on every run.

## Testing

- Drove the specs red before implementing. The "keeps a supplied due date" failure returned `certification_date + 30` instead of the supplied date, confirming the value was wiped and rebuilt rather than merely unset.
- Reintroduced the defect after implementing to confirm the request specs discriminate through the full HTTP path, including the `UnionObject` type dispatch that fires the callbacks.
- Reordered the two `before_validation` callbacks to confirm the ordering spec fails.
- `make test`: 3535 examples, 0 failures. Coverage 96.85% line, 84.09% branch.
- `make lint`: 584 files, no offenses.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/spec/models/certifications/requirement_params_spec.rb
+++ b/reporting-app/spec/models/certifications/requirement_params_spec.rb
@@ -208,6 +208,31 @@ RSpec.describe Certifications::RequirementParams do
       end
     end
 
+    context "when the request supplies a due date that is not a date" do
+      subject(:validity) { params.valid? }
+
+      [
+        [ "an array", [ "2026-12-15" ] ],
+        [ "an integer", 12_345 ],
+        [ "a boolean", true ]
+      ].each do |label, value|
+        context "when it is #{label}" do
+          let(:params) do
+            described_class.new_filtered(
+              "certification_date" => certification_date,
+              "certification_type" => "recertification",
+              "due_date" => value
+            )
+          end
+
+          it "rejects the request" do
+            expect(validity).to be false
+            expect(params.errors).to include(:due_date)
+          end
+        end
+      end
+    end
+
     context "when the request supplies neither a due date nor a due period" do
       let(:params) do
         build(
--- a/reporting-app/spec/models/certifications/requirement_params_spec.rb
+++ b/reporting-app/spec/models/certifications/requirement_params_spec.rb
@@ -117,4 +117,114 @@ RSpec.describe Certifications::RequirementParams do
       end
     end
   end
+
+  describe "due date" do
+    # The API reaches these callbacks through UnionObject.new, which calls valid?
+    # as its type dispatch. Validating here is what production actually does.
+    subject(:requirements) do
+      params.valid?
+      params.to_requirements
+    end
+
+    # Sits well away from today, so anchoring on the certification date is
+    # distinguishable from anchoring on the day the request is processed.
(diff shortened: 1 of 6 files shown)


reporting-app/spec/models/certifications/requirement_params_spec.rb
```diff
@@ -117,4 +117,114 @@
...
+      end
+
+      it "accepts the request and offsets by the default due period" do
+        expect(requirements.due_date).to eq Date.current + 30.days
+      end
+
+      it "no longer rejects the request for having neither field" do
+        expect(params).to be_valid
+      end
+    end
```
jdettmannnava, line 253: What if it supplies nonsense?

jdettmannnava (approved):
LGTM but a test for when non-date is supplied would make me happy

reporting-app/spec/models/certifications/requirement_params_spec.rb
giverm: In addition to the test, I added a check to ensure `due_date` is a date

PR #891, opened 2026-09-03
PBI-71033: Record the application date on every creation path

giverm:
## Ticket

Resolves PBI-71033

## Changes

- `UnifiedRecordProcessor#build_certification` sets `application_date` from
the file's certification date column
- `Demo::Certifications::CreateForm` collects `application_date` in place of
`certification_date`, and supplies `certification_date` to `RequirementParams`
explicitly
- The `:certification` factory sets an `application_date` default, covering the
demo path and `db/seeds/development.rb`
- `CertificationRequestPage` (e2e) locates the renamed form field

## Context for reviewers

The API already recorded `application_date`. Batch upload and the demo form did
not, so a certification created by either carried nothing in that column.
Nothing reads the field yet, so no determination outcome changes in this PR. It
is populated now because the exclusion repoint that follows needs an anchor on
every creation path, and a missing anchor does not raise, it silently stops
every exclusion firing.

## Testing

```
$ make lint
586 files inspected, no offenses detected

$ make test
3569 examples, 0 failures
Line coverage: 5713 / 5898 (96.86%)
Branch coverage: 1387 / 1651 (84.00%)

$ cd e2e && npm run type-check
(clean)
```

Manually created a certification through the demo form and confirmed the
persisted row carries a populated `application_date`.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/forms/demo/certifications/create_form.rb
+++ b/reporting-app/app/forms/demo/certifications/create_form.rb
@@ -29,8 +29,7 @@ module Demo
       end
 
       def to_certification
-        # RequirementParams slices to its own attribute names and still validates
-        # certification_date; supply it until that field is removed from the model.
+        # TODO: drop this merge when certification_date leaves the model, still validated present until all readers have been repointed.
         certification_requirement_params = ::Certifications::RequirementParams.new_filtered(
           attributes.with_indifferent_access.merge(certification_date: application_date)
         )
@@ -47,11 +46,6 @@ module Demo
           return false
         end
 
-        # Generated member data is anchored on the application date. That is correct only for a
-        # new application, where the evaluated month is the application month. Recertification
-        # and change in circumstance evaluate a month derived from the certification period,
-        # which this form does not collect, so their generated data stops lining up once the
-        # exclusion repoint makes the evaluated month type-dependent.
         member_data = {}
 
         case external_scenario
--- a/reporting-app/app/forms/demo/certifications/create_form.rb
+++ b/reporting-app/app/forms/demo/certifications/create_form.rb
@@ -29,7 +29,11 @@ module Demo
       end
 
       def to_certification
-        certification_requirement_params = ::Certifications::RequirementParams.new_filtered(attributes.with_indifferent_access)
+        # RequirementParams slices to its own attribute names and still validates
+        # certification_date; supply it until that field is removed from the model.
+        certification_requirement_params = ::Certifications::RequirementParams.new_filtered(
+          attributes.with_indifferent_access.merge(certification_date: application_date)
+        )
         # shouldn't be possible, but we need to ensure the params are valid in
         # order to construct the requirements next
         if certification_requirement_params.invalid?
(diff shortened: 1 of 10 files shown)


reporting-app/app/forms/demo/certifications/create_form.rb
```diff
@@ -43,78 +47,83 @@ def to_certification
           return false
         end
 
+        # Generated member data is anchored on the application date. That is correct only for a
```
jdettmannnava: Odd place for this comment; in fact, I am not sure it needs to be in the code at all?

reporting-app/app/forms/demo/certifications/create_form.rb
```diff
@@ -29,7 +29,11 @@ def self.new_for_certification_type(certification_type)
       end
 
       def to_certification
-        certification_requirement_params = ::Certifications::RequirementParams.new_filtered(attributes.with_indifferent_access)
+        # RequirementParams slices to its own attribute names and still validates
+        # certification_date; supply it until that field is removed from the model.
+        certification_requirement_params = ::Certifications::RequirementParams.new_filtered(
+          attributes.with_indifferent_access.merge(certification_date: application_date)
```
jdettmannnava, line 34: So, we are still storing this as certification date in the requirements?

reporting-app/app/services/unified_record_processor.rb
```diff
@@ -109,6 +109,10 @@ def build_certification(record)
     Certification.new(
       member_id: record["member_id"],
       case_number: record["case_number"],
+      # The file carries one date, so it is recorded as the application date. Exact for
```
jdettmannnava: Again, this comment implies the need for more work, but there isn't any direct call to action. Explaining why application_date == record['certification_date'] would be more valuable, but even that is a pretty common feature for code drift

reporting-app/spec/forms/demo/certifications/create_form_spec.rb
```diff
@@ -3,10 +3,28 @@
...
+    subject(:certification) { form.to_certification }
+
+    let(:collected) { Date.new(2026, 3, 15) }
+    let(:form) { described_class.new(application_date: collected) }
+
+    it "records the date collected on the form as the certification's application date" do
+      expect(certification.application_date).to eq(collected)
+    end
+
+    # Regression guard for the params slice: RequirementParams.new_filtered slices to that
```
jdettmannnava: The problem of regressions is interesting in our case, where we don't really have any production usage, and I am inclined to allow regressions in order to make the code cleaner.

jdettmannnava (approved):
It does what it says!

I tested the demo form, but batch upload failed on "undefined method 'storage_adapter' for an instance of Rails::Application::Configuration", and I didn't think it was worth debugging.

I am a bit concerned about leaving certification_date in the Requirements long-term.

reporting-app/app/services/unified_record_processor.rb
```diff
@@ -109,6 +109,10 @@ def build_certification(record)
...
+      # new_application rows and an approximation for recertification rows, which the
+      # published template cannot yet distinguish.
+      application_date: Date.parse(record["certification_date"]),
```
jgavin-nava: Is the Date.parse necessary since `BatchUploadRecordValidator` already validates that it is a date and if it somehow gets to here and is fails it would raise an error outside the `persist!`'s `RecordInvalid` rescue

jgavin-nava (approved):
LGTM as well. The other thing is we still allow for application date to be null on the api route if this is truly not allowing it maybe we should add a validation on that and change the open API as well.

reporting-app/app/services/unified_record_processor.rb
giverm: Good catch. I dropped it

reporting-app/app/forms/demo/certifications/create_form.rb
giverm: Agreed. It's been deleted

reporting-app/app/forms/demo/certifications/create_form.rb
giverm: For now. There are still have few uses of `certification_date` that need to be replaced. It will be dropped at the end.

giverm:
> LGTM as well. The other thing is we still allow for application date to be null on the api route if this is truly not allowing it maybe we should add a validation on that and change the open API as well.

Agreed, the validation is going to land in follow up work.

PR #893, opened 2026-09-04
PBI-70969 Create external activity

jdettmannnava:
## Ticket

Resolves #70969

## Changes

Adds  the `ExternalActivity` model that will consolidate `ExternalHourlyActivity` and
`ExternalIncomeActivity`, where a single row reports hours, gross income, or both.

- New `ExternalActivity` — union of both models' categories (including `household`) and source
  types (`api` and `batch_upload`), the shared `for_member` / `within_period` scopes, and new
  `with_hours` / `with_income` scopes plus `#hours?` / `#income?` for separating the two tracks.
- Hours are now validated against the **reported period** rather than a flat annual ceiling.
- `:external_activity` factory and model spec.
- Locale entries for the two new error messages and for the `batch_upload` source label.

Nothing reads or writes `ExternalActivity` yet — this is purely additive. The intake service and the
read-path cutover follow in separate PRs, and both old tables stay in place until a later backfill.

## Context for reviewers

Two smaller notes:

`household` stays external-only in `ExternalActivity::ALLOWED_CATEGORIES` rather than moving into
`ActivityCategories`, since a member-reported `Activity` can never be household income.

## Testing

Migration applied and rolled back cleanly, and `make db-reset` confirms the check constraint
round-trips through `schema.rb` (which is what `db-reset` loads, not the migrations):

```
== 20260903145559 CreateExternalActivities: migrating ====
-- create_table(:external_activities, ...)
== 20260903145559 CreateExternalActivities: migrated ====

t.check_constraint "hours IS NOT NULL OR gross_income IS NOT NULL", name: "external_activities_hours_or_income"
t.index ["period_start", "period_end"], name: "index_external_activities_on_period"
t.index ["source_type", "source_id"], name: "index_external_activities_on_source"
```

`make lint` — 580 files inspected, no offenses.

Full suite:

```
3614 examples, 0 failures

Line coverage: 5755 / 5939 (96.90%)
Branch coverage: 1396 / 1659 (84.14%)
```

The model spec (51 examples) covers each value combination — hours-only, income-only, both, and
neither rejected by the model *and* by the DB constraint when validations are bypassed — the hours
ceiling at its boundaries (`days * 24` accepted, 25 hours in a one-day period and 500 in a one-week
period rejected, income unconstrained, reversed periods reported once), `within_period` containment
including the end-of-month widening, and that a combined row appears in **both** `with_hours` and
`with_income` while chaining them selects only combined rows.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/spec/models/external_activity_spec.rb
+++ b/reporting-app/spec/models/external_activity_spec.rb
@@ -1,0 +1,381 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ExternalActivity, type: :model do
+  describe 'factory' do
+    it 'creates a valid hours-only record' do
+      expect(build(:external_activity, :with_hours)).to be_valid
+    end
+
+    it 'creates a valid income-only record' do
+      expect(build(:external_activity, :with_income)).to be_valid
+    end
+
+    it 'creates a valid record carrying both hours and income' do
+      expect(build(:external_activity, :with_hours_and_income)).to be_valid
+    end
+
+    it 'creates a valid household income record' do
+      expect(build(:external_activity, :household)).to be_valid
+    end
+  end
+
+  describe 'validations' do
+    subject(:activity) { build(:external_activity, :with_hours) }
+
+    describe 'member_id' do
+      it 'is required' do
+        activity.member_id = nil
+        expect(activity).not_to be_valid
+        expect(activity.errors[:member_id]).to include("can't be blank")
+      end
+    end
+
+    describe 'category' do
+      it 'is required' do
+        activity.category = nil
+        expect(activity).not_to be_valid
+        expect(activity.errors[:category]).to include("can't be blank")
+      end
+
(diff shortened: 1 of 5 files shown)


reporting-app/spec/models/external_activity_spec.rb
```diff
@@ -0,0 +1,381 @@
...
+      end
+
+      it 'accepts every allowed category' do
+        ExternalActivity::ALLOWED_CATEGORIES.each do |category|
+          activity.category = category
+          expect(activity).to be_valid
+        end
+      end
+
+      it 'accepts the household category' do
```
giverm, line 49: Should we also validate that income must be present when the category is `household`?

giverm (approved):

reporting-app/spec/models/external_activity_spec.rb
jdettmannnava: This is handled at the API layer, where the gross income has to be greater than 0

PR #894, opened 2026-09-04
PBI-70969: Add external activity service

jdettmannnava:
## Ticket

Part of PBI-70969

## Changes

Adds `ExternalActivityService`, the single intake path for `ExternalActivity`, superseding
`ExternalHourlyActivityService` and `ExternalIncomeActivityService`. One submission may report hours,
gross income, or both, and is split into one entry per calendar month it touches.

- `ExternalActivityService` — `create_entries` / `create_entry` / `duplicate_entry?` taking `hours:`
  and `gross_income:` together, one fingerprint covering both values, every row written inside
  `Strata::AuditLog` under a single `external_activity.create` action, and one
  `recalculate_compliance:` keyword in place of the old per-track flags.
- `ActivityAggregator#apportioned_multi_values_map` — apportions several values across one set of
  month periods using the same weights. The cumulative-share arithmetic and weight selection are
  extracted so the single- and multi-value paths share them; `daily_values_map` and
  `monthly_values_map` keep their signatures.
- Compliance recalculation records **exactly one** determination per save: hours first, falling
  through to income only when hours fall short.
- Merged service spec (86 examples, replacing the two 391- and 589-line specs at cutover) and new
  `apportioned_multi_values_map` coverage.

Nothing calls `ExternalActivityService` yet — `Certifications::CreationService` still uses the old
pair, and the old services are untouched. This is purely additive, like the model in #70969's first
PR. The read-path and intake cutover follow in the next PR.

## Context for reviewers

`apportioned_multi_values_map` apportions each value over one shared set of month periods and keeps a
month when **any** value has a non-zero share; a value whose own share rounded away is left `nil` for
that month and rolled into a later one by the running total. `nil` rather than zero matters: the model
validates `greater_than: 0` with `allow_nil`, so a zero share would fail the save.

**Which split rule applies** is decided once per submission, from whether hours are present:

- hours present → both columns apportion **by days**
- income only → the existing income rule (evenly across whole calendar months, by days otherwise)

So adding hours to a payload changes how its income is apportioned. That falls out of keeping one
activity's two halves on the same boundaries, but it will look surprising in isolation.

**One determination, hours first.** Both determination services' `.calculate` methods aggregate *and*
record, so calling both for a combined row would write two `Determination` rows — and
`Determination.latest_per_subject` orders by `created_at`, so which one drove member status would
depend on insertion order. Instead `record_one_determination` decides the branch and records only the
winning track, building on the services' already-public `aggregate_*` and `compliant_for_*?` methods.

The open certification is resolved **once**, before either track: `open_certification_id_for_member`
filters on open cases and recording a compliant outcome closes the case, so a per-track lookup would
let a compliant close silently skip the second track.

Hours compliance is judged on monthly hours alone, matching
`HoursComplianceDeterminationService#calculate` — the education-enrollment track belongs to
`CommunityEngagementCheckService`, so consulting it here would quietly widen this outcome.

**Why the recalculation specs stub the threshold checks.** `ActivityAggregator` still fetches from the
superseded tables in this PR, so a freshly written `ExternalActivity` row does not appear in the
aggregates yet. The aggregates and both record paths run for real — real determination rows, real case
closure — but the compliant/not-compliant branch is driven by stubbing
`compliant_for_monthly_hours?` / `compliant_for_monthly_income?`. End-to-end figures become testable in
the cutover PR, once reads move to `external_activities`.

One behavioural note carried over from the schema: hours-bearing rows now get a `reported_at` (the
hours table had no such column) and can carry `metadata["employer"]`.

## Testing

`make lint` — 582 files inspected, no offenses.

Full suite:

```
3708 examples, 0 failures

Line coverage: 5826 / 6012 (96.90%)
Branch coverage: 1429 / 1698 (84.15%)
```

Up 94 examples from the previous PR, with coverage holding.

The merged service spec ports the full behaviour of both old specs — day-proportional hours splits
across partial months and year boundaries, even income splits for whole months, rounding that sums
back to the total without drift, `source_id` / `name` / `origin_hash` stamped across a submission,
reversed periods and blank values reaching the model as-is, and the duplicate rules including
household identity by tax ID and date of birth — then adds:

- Combined submissions: one entry per month carrying both values, both totals apportioned in full, and
  income apportioned by days rather than evenly because hours are present.
- Month alignment when one value's share rounds to zero: 300 hours with `0.02` income over
  Jan 31 – Mar 31 keeps all three months' hours, leaves income unset in January, and still sums both
  totals exactly.
- Fingerprint distinctness across the three shapes — the same total as income rather than hours, and
  adding income to an existing hours submission, are both new submissions rather than duplicates.
- All six recalculation branches, including that a combined row whose hours qualify records **only**
  the hours determination and never consults income, and that a combined row whose hours fall short
  records only the income one.
- Recalculation runs once per submission rather than once per monthly entry, and not at all for a
  duplicate that created nothing.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/services/external_activity_service.rb
+++ b/reporting-app/app/services/external_activity_service.rb
@@ -1,0 +1,173 @@
+# frozen_string_literal: true
+
+# Service for creating and validating ExternalActivity entries from API and batch intake.
+# Supersedes ExternalHourlyActivityService and ExternalIncomeActivityService: one submission may
+# report hours, gross income, or both, and is split into one entry per calendar month it touches.
+#
+# After a successful save, optional compliance recalculation (+recalculate_compliance+, default
+# +true+) records exactly one determination for the member's open case — hours first, falling
+# through to income only when hours fall short. Certification intake passes
+# +recalculate_compliance: false+ so rows created before the case exists do not run this path.
+class ExternalActivityService
+  include Strata::VirtualActor
+
+  AUDIT_ACTION_CREATE = "external_activity.create"
+
+  class << self
+    include ActivityAggregator
+    include OriginHash
+
+    # Create one entry per calendar month the period touches.
+    # A submission already on file is logged and skipped, leaving the caller's other work intact.
+    # @param identity [Array] values beyond the name identifying whose figures these are (household
+    #   income passes the member's tax ID and date of birth), folded into the fingerprint.
+    # @return [Array<ExternalActivity>] the entries created, empty for a duplicate
+    # @raise [ActiveRecord::RecordInvalid] on validation failure
+    def create_entries(member_id:, category:, period_start:, period_end:, source_type:,
+                       hours: nil, gross_income: nil, source_id: nil, reported_at: Time.current,
+                       metadata: {}, name: nil, employer: nil, identity: [],
+                       recalculate_compliance: true)
+      origin_hash = origin_hash_for(member_id, category, hours, gross_income,
+                                    period_start, period_end, name, *identity)
+
+      if duplicate_entry?(origin_hash)
+        log_duplicate_submission(origin_hash, member_id:, category:, period_start:, period_end:)
+        return []
+      end
+
+      entries = month_entries(hours:, gross_income:, period_start:, period_end:)
+        .map do |current_period_start, current_period_end, values|
+          create_entry(member_id:, category:,
+                       hours: values[:hours], gross_income: values[:gross_income],
+                       period_start: current_period_start, period_end: current_period_end,
+                       source_type:, source_id:, reported_at:, metadata:, name:, employer:,
+                       origin_hash:, recalculate_compliance: false)
+        end
+
+      maybe_recalculate_compliance(member_id, hours:, gross_income:) if recalculate_compliance
+      entries
+    end
+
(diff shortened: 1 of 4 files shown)


reporting-app/app/services/external_activity_service.rb
```diff
@@ -0,0 +1,173 @@
...
+      ExternalActivity.exists?(origin_hash: origin_hash)
+    end
+
+    private
+
+    # Hours present means both values apportion by days, so the two halves of one activity divide
+    # along the same month boundaries. An income-only submission keeps the income rule: evenly
+    # across whole calendar months, by days otherwise.
+    def month_entries(hours:, gross_income:, period_start:, period_end:)
+      if hours.present?
```
giverm: Should `hours = 0, gross_income = 580` be considered and income-only submission? Currently, this would split with a daily weight in that case. (Please ignore if this seems too nit-picky)

giverm (approved):
Looks great!

reporting-app/app/services/external_activity_service.rb
jdettmannnava: Good idea!

PR #895, opened 2026-09-04
PBI-70969: switch to single external activity service

jdettmannnava:
## Ticket

Part of PBI-70969

## Changes

The cutover. Every read and write moves to `ExternalActivity`, and the two superseded models become
read-only stubs. Net **574 insertions, 2,160 deletions** across 50 files.

- `ActivityAggregator` — both fetch helpers now read the one table, with new `with_hours` /
  `with_income` scopes doing the separating. `CertificationCasesController`,
  `MemberDashboardCompliance` and `MemberDashboardComplianceService` follow.
- `Certifications::MemberData::Activity` — drops the `type` discriminator; which values are present
  decides what is recorded. Adds `ALLOWED_SOURCES`.
- `Certifications::CreationService` — one pass over `member_data.activities` instead of separate
  hourly and income passes, so an activity reporting both becomes a single row per month.
- `ExternalHourlyActivity` / `ExternalIncomeActivity` — reduced to read-only stubs (`readonly?`
  returns `true`) over their retained tables. Both intake services, both factories and their four
  specs are deleted.
- OpenAPI — the `oneOf` of hourly and income activity schemas collapses to one object.
- `Determinations::HoursBasedDeterminationData` / `IncomeBasedDeterminationData` — unchanged keys,
  with a note on the id referent.
- Supersedes the ADR in `docs/architecture/income-data/income-data.md` that chose parallel tables.

## Context for reviewers

**Employer is now recorded on hours-bearing rows** (in `metadata`, as it already was for income).
Dropping it only when hours are present is arbitrary once one row carries both, so the spec that
asserted the old behaviour is inverted.

- Both old tables are retained, so rows written before the cutover stop counting toward compliance
  until the backfill task runs. Fine for dev/sandbox; there is no production environment.

**One thing for product:** a row carrying both values now appears in *both* the "Hours reported" and
"Income reported" tables, on the staff case page and the member dashboard. It genuinely contributes
to both compliance tracks, but it is the same activity shown twice.

## Testing

`make lint` — 576 files inspected, no offenses. Full suite:

```
3571 examples, 0 failures

Line coverage: 5723 / 5917 (96.72%)
Branch coverage: 1406 / 1672 (84.09%)
```

Tested via api with three activities: 1 income-only, 1 hours-only, and 1 income and hours. Varied the amounts to produce compliant via income, compliant via hours, compliant via both, and not compliant.

Staff view:
Screenshot 2026-09-04 at 2 09 15 PM

♻️ Environment destroyed ♻️


reporting-app/app/services/external_income_activity_service.rb
```diff
@@ -1,116 +0,0 @@
...
-
-      if duplicate_entry?(origin_hash)
-        log_duplicate_submission(origin_hash, member_id:, category:, period_start:, period_end:)
-        return []
-      end
-
-      month_values = if whole_months?(period_start, period_end)
-        monthly_values_map(period_start, period_end, gross_income)
-      else
-        daily_values_map(period_start, period_end, gross_income)
```
giverm, line 37: `monthly_values_map` and `daily_values_map` and now unused and can be removed (along with `apportioned_values_map`)

giverm (approved):
LGTM! 🚀

PR #897, opened 2026-09-04
PBI-71081: evaluate exclusions from the application date

giverm:
## Ticket

Resolves PBI-71081

## Changes

- `Certification#evaluated_month` returns `application_date.beginning_of_month`,
the single resolver for the month a member's exclusion status is evaluated for.
- `ExclusionDeterminationService` sets the rules-engine fact `evaluated_month`
from that resolver, replacing the `certification_date` it read off
`certification_requirements`.
- `Rules::ExclusionRuleset` renames the `certification_date` parameter to
`evaluated_month` across the eight date-based rules and `meets_end_condition`.
The comparison bodies are unchanged; the `cert_month` locals become `month`,
and the comments naming the certification date as the anchor are corrected.

## Context for reviewers

`certification_date` is being removed from the certifications API contract, so
the exclusion rules need a different anchor. They now take their month from
`application_date`. The demo tooling and batch upload write the same value to
both fields, and nothing in `app/`, `lib/` or `spec/` compares the two, so no
member's outcome changes.

`evaluated_month` is a method rather than an inline expression at the one call
site because it names a domain concept. The field it replaces was answering two
different questions, which month a member's status is evaluated for and which
months they must report activities on, and separating those is the point of this
ticket set. It sits on `Certification` because that is the only object holding
both candidate fields: `application_date` is a column, `certification_period_start`
is inside the `certification_requirements` jsonb. It will need updating when
recertification logic is introduced.

The rules engine binds facts by parameter name, so the fact key and the nine
parameters had to move together. A mismatch is silent: it resolves to nil, every
rule bails, and the member is reported as not excluded with nothing failing. The
rule bodies themselves needed no edits, since each already calls
`.beginning_of_month`, which a month-start Date makes idempotent.

## Testing

`make lint` clean (586 files, no offenses) and `make test` green (3571 examples,
0 failures; line 96.86%, branch 84.02%).

Coverage is two resolver specs plus one behavioural example, in which the
application date and the certification date name different months and only the
application month falls inside the exemption period.

Both were checked against mutated code rather than assumed. With the service
still feeding the certification date and the resolver left correct, that
behavioural example is the only one of 3571 that fails, so it is what proves the
repoint is wired rather than merely present. With the resolver returning the
certification date instead, the two resolver specs fail alongside it.

The existing exclusion fixtures had diverged: every exemption period is built
around `cert_date` of 2025-07-01 while the factory defaults `application_date`
to today, and the spec set the field zero times. Both `let` blocks now pin it
explicitly, rather than deriving it in the factory, so the diff shows which
fixtures depended on the anchor. Removing either pin fails loudly (11 examples
and 1).

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/certification.rb
+++ b/reporting-app/app/models/certification.rb
@@ -91,4 +91,11 @@ class Certification < ApplicationRecord
   def region
     certification_requirements&.region
   end
+
+  # The month a member's exclusion status is evaluated for. Bound into the rules
+  # engine as the `evaluated_month` fact, which resolves by matching the ruleset's
+  # parameter name -- renaming one side without the other fails silently.
+  def evaluated_month
+    application_date&.beginning_of_month
+  end
 end
--- a/reporting-app/app/models/rules/exclusion_ruleset.rb
+++ b/reporting-app/app/models/rules/exclusion_ruleset.rb
@@ -12,71 +12,71 @@ module Rules
     # Incarceration excludes through this many months after the incarceration month
     INMATE_BUFFER_MONTHS = 3
 
-    def is_pregnant(pregnancy, postpartum, certification_date)
-      meets_end_condition(pregnancy, certification_date) || meets_end_condition(postpartum, certification_date)
+    def is_pregnant(pregnancy, postpartum, evaluated_month)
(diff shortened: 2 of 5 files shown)


jdettmannnava (approved):
LGTM

PR #909, opened 2026-09-09
PBI-71083: derive the reportable months from the application date

giverm:
## Ticket

Resolves PBI-71083

## Changes

- `Certifications::RequirementParams#months_that_can_be_certified` takes the
anchor as an argument and counts back from the application date rather than
`certification_date`. `#to_requirements` gains a required `application_date:`
keyword to pass it through. The arithmetic is unchanged.
- `Api::Certifications::CreateRequest` validates `application_date` present and
threads it into `to_requirements`.
- `CertificationService#certification_requirements_from_input` gains a required
`application_date:` keyword. `UnifiedRecordProcessor` passes the value it
already writes to the column, cast explicitly because the CSV holds dates as
strings.
- `Demo::Certifications::CreateForm` passes the anchor it already holds.
- `Certification.find_duplicate` keeps its blank guard, with a TODO on the
`application_date` clause this makes unreachable from the API.
- `oas.json` and `openapi.generated.yml` mark `application_date` required and
non-nullable, and all three request examples carry it.

## Context for reviewers

The anchor is a method argument rather than an attribute on `RequirementParams`,
which is load-bearing. That class is a union member and `UnionObject#new` returns
the first member that is `valid?`, so any presence validation on it is also type
dispatch. The application date is not in the nested requirements hash, so
validating it there would match neither member and reject every request.
`CreateRequest` validates it instead.

Deriving a recertification's months from its certification period stays out of
scope, deferred until the certification period contract is defined. Every type still
counts backwards,
so this is behaviour-preserving. The derivation carries a TODO.

Two API specs flipped from 201 to 422. They asserted that an undated request
bypassed duplicate detection, which the API can no longer express;
`find_duplicate` keeps its coverage at the model level.

## Testing

`make lint` clean (586 files, no offenses) and `make test` green (3589 examples,
0 failures; line 96.86%, branch 84.22%).

Coverage is the months following the application date when the two dates differ,
asserted at unit, service and request level; a request with no application date
rejected naming the field; and a nested `application_date` in the requirements
hash ignored rather than becoming the anchor. The service-level example compares
two real requests rather than a literal, so it cannot drift as the factories
change.

Running it caught what reading did not: the batch path passed the raw CSV string,
since moving the value out of an attribute moved it out of the date cast it had
been getting for free.

No migration, so E2E is a real signal here rather than the usual migration-PR
noise.

♻️ Environment destroyed ♻️

### DIFF
--- a/reporting-app/app/models/api/certifications/create_request.rb
+++ b/reporting-app/app/models/api/certifications/create_request.rb
@@ -10,9 +10,12 @@ class Api::Certifications::CreateRequest < ValueObject
   attribute :household_data, Certifications::HouseholdData.to_type
 
   validates :certification_requirements, presence: true
-  # Not on RequirementParams: it is a union member whose valid? decides type dispatch, and this
-  # field is not in the nested requirements hash. See RequirementParams#to_requirements.
+  # application_date is a Certification attribute, not a RequirementParams one: RequirementParams'
+  # valid? doubles as type dispatch (UnionObject#new), so presence-validating application_date
+  # there would make every parameter-shaped request invalid, matching neither union member. See
+  # RequirementParams#to_requirements.
   validates :application_date, presence: true
+  validate :application_date_must_be_a_date
 
   def self.from_request_params(params)
     new_filtered(params)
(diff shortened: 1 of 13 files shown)


reporting-app/app/models/api/certifications/create_request.rb
```diff
@@ -10,6 +10,12 @@ class Api::Certifications::CreateRequest < ValueObject
   attribute :household_data, Certifications::HouseholdData.to_type
 
   validates :certification_requirements, presence: true
+  # application_date is a Certification attribute, not a RequirementParams one: RequirementParams'
+  # valid? doubles as type dispatch (UnionObject#new), so presence-validating application_date
+  # there would make every parameter-shaped request invalid, matching neither union member. See
+  # RequirementParams#to_requirements.
```
jgavin-nava: From past reviews I think this probably an unnecessary comment. It makes more sense to me just leave the call out in the PR description and remove it from the code. It seems obvious for the reasoning and it is explained more in the to_requirements section.

jgavin-nava (approved):
LGTM otherwise

PR #914, opened 2026-09-14
PBI-71253: name the evaluated month on the member dashboard

giverm:
## Ticket

Part of PBI-71253

## Changes

The member dashboard stops reading the certification date. Six files, +28/-12.

- `MemberDashboardCompliance` — the `certification_date` attribute becomes
  `evaluated_month`, across its reader, initializer and documentation.
- `MemberDashboardComplianceService` — supplies `Certification#evaluated_month`
  rather than reaching into `certification_requirements`.
- `MemberComplianceHelper` and `DashboardController` — read the renamed
  attribute.

## Context for reviewers

Two member-facing surfaces name a month from this value: the activity report
heading, and the "Current period" header. Both now name the month OSCER
evaluates, which is what decided the member's status.

`evaluated_month` is the method the exclusion path already uses, bound into the
rules engine as the fact of the same name. Reading it here means the member sees
the month that decided their status rather than a second derivation that could
drift from it.

The value object has one construction site, so the keyword-argument change has
no other callers.

## Testing

```
3623 examples, 0 failures

Line coverage: 5773 / 5959 (96.87%)
Branch coverage: 1439 / 1706 (84.34%)
```

A new service-spec example pins the behaviour against a certification whose
application date and stored certification date fall in different months.

Drew the dashboard in the running app for a member whose application date is
11 September 2026 and whose stored certification date is 7 October 2026. The
heading reads "September 2026 Activity Report", so the rendered value
discriminates between the two fields rather than only matching a label.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

♻️ Environment destroyed ♻️


jgavin-nava (approved):
LGTM

PR #922, opened 2026-09-15
PBI-71253: repoint batch intake to the application date

giverm:
## Ticket

Part of PBI-71253

## Changes

Batch CSV intake stops reading the certification date. 15 files, +68/-150.

- `UnifiedRecordProcessor` — `REQUIRED_FIELDS` names `application_date`; the duplicate check, the certification's anchor and the pregnancy placeholder read that column.
- `Certification` — `exists_for?` deleted. `find_duplicate` is the only duplicate check now, and its comment records why the blank guard stays.
- `BatchUploadRecordValidator` — `DATE_FIELDS` follows the rename; `REQUIRED_FIELDS` already delegates.
- Staff upload page, its locale file, the downloadable template and the sample CSV name the new column.
- `batch-upload-instructions.md` and `README_ERRORS.md` — required-field table, example header, and the VAL_001 / DUP_001 field lists.

## Context for reviewers

**Breaking change for any existing CSV.** A `certification_date` header now fails VAL_001 with "Missing required fields: application_date". The template, sample CSV and both docs move with it.

**Consolidated onto `find_duplicate`, which already asked the same question.** `exists_for?` checked against a different anchor than the API's check, so the two paths could disagree: an API row created with a differing certification date could not be matched by a later upload of the same member, case and date.

One `certification_date` key stays, feeding `requirement_input`, because `RequirementParams` still presence-validates it. TODO in place; it goes with the final removal alongside the demo form.

**Known gaps, out of scope:** nothing validates the header row, so an old-header file errors per row rather than once; and no unique constraint backs the duplicate key, so parallel chunks can both pass the check.

## Testing

```
3625 examples, 0 failures

Line coverage: 5772 / 5957 (96.89%)
Branch coverage: 1440 / 1706 (84.40%)
```

Three uploads through the staff page against the running app: the new header accepted, the same file re-uploaded caught as DUP_001 with no second row written, and the old header rejected as VAL_001. Dev DB restored to its 21-certification baseline.

♻️ Environment destroyed ♻️


jdettmannnava (approved):
LGTM.
Also successfully uploaded

PR #923, opened 2026-09-16
PBI-71253: widen the contract and repoint the factory anchor

giverm:
## Ticket

Part of PBI-71253

## Changes

Preparation for the certification date removal. No production behavior changes.
5 files, +13/-11. Bottom of a two-PR stack; the removal itself follows on top.

- `certification_requirements_factory` derives `months_that_can_be_certified` from a new
`application_date` transient rather than `certification_date`. `certifications_factory`
passes its own value down, so a built certification's months match what production would
derive for it.
- `oas.json` and `openapi.generated.yml` drop `certification_date` from all three
`required` arrays. `CertificationRequirements` requires `months_that_can_be_certified`
in its place.
- `requirements_spec` now asserts the client's months survive dispatch, not just the
returned class.

## Context for reviewers

**Split so that a red suite in the follow-up means the removal.** Deleting the model
attribute while the contract still requires it turns the suite red for a schema reason:
`rails_helper` parses `openapi.generated.yml` into `OPENAPI_DOC` and the API request spec
validates responses against it 39 times. Widening the contract first keeps that signal
diagnostic.

## Testing

Full suite green at 3625 examples, RuboCop clean across 586 files.

The contract edit is the load-bearing part, and `spec/requests/api/certifications_spec.rb`
is the fastest signal on it: 39 `match_openapi_doc` assertions run through that file, so a
wrong `required` array turns it red immediately rather than surfacing later.

The factory repoint reaches beyond specs. `lib/dev/member_dashboard_frame_setup.rb` builds
certifications through the same factory without overriding requirements, so its months move
from a random anchor onto the certification's real application date.

♻️ Environment destroyed ♻️


reporting-app/lib/assets/oas.json
```diff
@@ -221,7 +221,7 @@
 	  }
         },
         "required": [
-          "certification_date"
+          "months_that_can_be_certified"
```
jdettmannnava: This doesn't seem to map to the API we will be using?

giverm: You are correct, and I expect this branch of the API schema will ultimately be removed as we bring things inline with Tomas's tech spec. I was trying to keep the three branches of this oneOf requirement valid, but looking at it now, it's just adding lines that will need to be removed any ways. Removing this now

jdettmannnava (approved):
LGTM!

PR #924, opened 2026-09-16
PBI-71253: remove the certification date from the contract and model

giverm:
## Ticket

Resolves PBI-71253

## Changes

The deletion itself. Stacked on #923, which widened the contract and repointed the
factory anchor; review that one first.

- `requirements.rb` and `requirement_params.rb` lose the attribute, its presence
validation, and the `to_requirements` pass-through.
- `unified_record_processor` and the demo create form lose the two shims that copied
the application date into the field.
- Both contract files lose three property definitions and four examples. The response
schema gains `application_date`, which it already emitted but never documented.
- Both union members now carry date type guards; only `RequirementParams` had them.
- About 60 spec references swept, including 12 prose ones that no identifier grep reaches.

## Context for reviewers

**Both presence validations had to go in one commit.** Deleted alone, either one gives a
months-bearing payload with no certification date the opposite semantics. Routed to
`Requirements` the client's months are honoured; routed to `RequirementParams` they are
silently discarded and re-derived, because `months_that_can_be_certified` is a method
there rather than an attribute. Nothing in the response says which branch ran.

**The type guards are deliberate added scope.** Validation on these classes doubles as
union dispatch, so removing the second presence check widened which payloads reach the
`Requirements` branch, and that branch had no type guard on its date fields. `["garbage"]`
casts to `[nil]`, which is `present?`, so it persisted and then broke the member's own
certification page on every load.

## Testing

Full suite green at 3627 examples, RuboCop clean across 586 files, and a codebase search
for the field returns nothing across app, lib, config, db, spec, public and the generated
spec.

Exercised against a running app rather than the suite alone:

- `POST /api/certifications` with a requirements-shaped payload returns 201 and keeps the
client's months, with `params` null confirming the branch.
- The same endpoint with a params-shaped payload returns 201 and derives the months, with
the `params` blob present.
- The demo form saves and determines, and the case closes on a month-dependent exclusion
rule rather than a presence-only one.
- A batch CSV imports once, and an identical re-upload is refused as a duplicate.

♻️ Environment destroyed ♻️


jgavin-nava (approved):
LGTM

